PR #30, opened 2020-03-02
Add dynamic list of goods

AlexanderShpilka:
https://alexandershpilka.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,58 @@
 import React from 'react';
+import { GoodsList } from './components/GoodsList';
+import { Button } from './components/Button';
 import './App.css';
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+const URL
+  = 'https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json';
 
-export default App;
+interface AppState {
+  goods: Good[];
+}
+
+async function getGoods(url: string): Promise<Good[]> {
+  const response = await fetch(url);
+
+  return response.json();
+}
+
+export class App extends React.Component<{}, AppState> {
+  state = {
+    goods: [],
+  };
+
+  handleLoadAllButtonClick = async () => {
+    const goods = await getGoods(URL);
+
+    this.setState({ goods });
+  };
+
+  handleLoad5ButtonClick = async () => {
+    const goods = await getGoods(URL);
+
+    this.setState({
+      goods: goods.slice(0, 5),
+    });
+  };
+
+  handleLoadRedButtonClick = async () => {
+    const goods = await getGoods(URL);
+
+    this.setState({
+      goods: goods.filter((good: Good) => good.color === 'red'),
+    });
+  };
+
+  render(): JSX.Element {
+    const { goods } = this.state;
+
+    return (
+      <div className="App">
+        <Button text="Load All" handler={this.handleLoadAllButtonClick} />
+        <Button text="Load 5" handler={this.handleLoad5ButtonClick} />
+        <Button text="Load Red" handler={this.handleLoadRedButtonClick} />
+        <GoodsList goods={goods} />
+      </div>
+    );
+  }
+}
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+
+interface ButtonProps {
+  text: string;
+  handler(): void;
+}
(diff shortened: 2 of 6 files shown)


src/App.tsx
```diff
@@ -1,8 +1,58 @@
 import React from 'react';
+import { GoodsList } from './components/GoodsList';
+import { Button } from './components/Button';
 import './App.css';
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+const URL
+  = 'https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json';
```
YuriiHoliuk: do not break a line just before/after =

src/App.tsx
```diff
@@ -1,8 +1,58 @@
 import React from 'react';
+import { GoodsList } from './components/GoodsList';
+import { Button } from './components/Button';
 import './App.css';
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+const URL
```
YuriiHoliuk: which url?
will be great to move to the constants?

src/App.tsx
```diff
@@ -1,8 +1,58 @@
...
+  = 'https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json';
 
-export default App;
+interface AppState {
+  goods: Good[];
+}
+
+async function getGoods(url: string): Promise<Good[]> {
```
YuriiHoliuk: if it is getGoods function - it shouldn't accept URL. If it accepts URL it should be named getFromApi | loadData etc.

src/App.tsx
```diff
@@ -1,8 +1,58 @@
...
+    this.setState({
+      goods: goods.slice(0, 5),
+    });
+  };
+
+  handleLoadRedButtonClick = async () => {
+    const goods = await getGoods(URL);
+
+    this.setState({
+      goods: goods.filter((good: Good) => good.color === 'red'),
```
YuriiHoliuk: it is a good practice, leave it as is but FYI, do not require to specify Good type here, because you are mapping field from typed state and typescript know which type is here

src/App.tsx
```diff
@@ -1,8 +1,58 @@
...
+    });
+  };
+
+  render(): JSX.Element {
```
YuriiHoliuk: But here will be better to remove `: JSX.Element`, because you have Component class and what if you will want to return `JSX.Element[]` or `string` from render?

src/App.tsx
```diff
@@ -1,8 +1,58 @@
...
+    const { goods } = this.state;
+
+    return (
+      <div className="App">
+        <Button text="Load All" handler={this.handleLoadAllButtonClick} />
```
YuriiHoliuk: instead of handler will be better to name it onClick, but ok

src/components/Button.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+interface ButtonProps {
+  text: string;
+  handler(): void;
+}
+
+export const Button = (props: ButtonProps): JSX.Element => {
```
YuriiHoliuk: use `React.FC` instead of `props: ButtonProps` and `: JSX.Element`

src/components/GoodItem.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+interface GoodItemProps {
+  name: string;
+  color: string;
+}
+
+export const GoodItem = (props: GoodItemProps): JSX.Element => {
```
YuriiHoliuk: the same as above

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { GoodItem } from './GoodItem';
+
+interface GoodsListProps {
+  goods: Good[];
+}
+
+export const GoodsList = (props: GoodsListProps): JSX.Element => {
```
YuriiHoliuk: ^

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { GoodItem } from './GoodItem';
+
+interface GoodsListProps {
```
YuriiHoliuk: I prefer to name it just Props(it is easier to always work with the same name), but your approach is also good, especially if you need to export this interface

PR #53, opened 2020-05-07
Develop

OlEzhA0:
[LINK](https://olezha0.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/getElement.tsx
+++ b/src/api/getElement.tsx
@@ -1,0 +1,12 @@
+const getData = async (URL: string) => {
+  const goods = await fetch(URL);
+  const list = await goods.json();
+
+  return list;
+};
+
+export const GetElement = async () => {
+  const URL = 'https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json';
+
+  return getData(URL);
+};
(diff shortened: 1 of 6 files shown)


src/api/getElement.tsx
```diff
@@ -0,0 +1,12 @@
+const getData = async (URL: string) => {
+  const goods = await fetch(URL);
+  const list = await goods.json();
+
+  return list;
+};
+
+export const GetElement = async () => {
```
irenhh, line 8: variable names start with lowercase letter

src/api/getElement.tsx
```diff
@@ -0,0 +1,12 @@
...
+  const URL = 'https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json';
```
irenhh, line 9: this url is usually stored as a global variable and exported wherever you need

src/components/Buttons.tsx
```diff
@@ -0,0 +1,36 @@
...
+      >
+        Load 5 first goods
+      </button>
+      <button
+        className="button"
+        type="button"
+        onClick={() => loadRedGoods()}
+      >
+        Load red goods
+      </button>
```
irenhh, line 31: would be great if you create an array that contains all info about each button and then you map that array and write button only once (since all buttons have the same structure)

PR #143, opened 2020-09-09
solution

Andrii256:
https://andrii256.github.io/react_dynamic-list-of-goods/

image

### DIFF
--- a/src/api/goods.js
+++ b/src/api/goods.js
@@ -1,11 +1,12 @@
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll() {
+// this is my personal compendium for future
+/*
+export function getGoodsFromServer() {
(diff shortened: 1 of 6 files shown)


oggyrock (approved):

irenhh:
you can't set default props for a nested value
you can read here more https://stackoverflow.com/questions/38123498/how-do-you-provide-default-props-for-nested-shape-in-react

src/api/goods.js
```diff
@@ -1,11 +1,12 @@
...
+export function getGoodsFromServer() {
   return fetch(API_URL)
     .then(response => response.json());
 }
+*/
 
-export const get5First = () => {};
-
-export const getRedGoods = () => {};
+export const getGoods = async() => (await fetch(API_URL)).json();
```
irenhh, line 12: if you return it directly, you don't need to write `await`

irenhh (approved):

PR #222, opened 2021-01-25
Solution

vovador:
[DEMO LINK](https://vovador.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,12 +2,57 @@ import React from 'react';
 
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { GoodList } from './GoodList';
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+class App extends React.Component {
+  state = {
+    goods: null,
+  }
+
+  getAllHandler = async() => {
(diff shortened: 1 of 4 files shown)


Serhii-Alieksieievych (approved):

aionova (changes requested):
Screenshot 2021-01-26 at 10 08 25

You need to render name of the goods, not the color. Color needs to be used in style. 
Also in get5First function goods need to be sorted, and only after that sliced

src/api/goods.js
```diff
@@ -6,6 +6,14 @@ export function getAll() {
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5First = async() => {
+  const goods = [...await getAll()];
 
-export const getRedGoods = () => {};
+  return goods.slice(0, 5);
```
aionova, line 12: You had to sort them first. Read the instructions

src/App.js
```diff
@@ -2,12 +2,57 @@ import React from 'react';
...
+    });
+  };
+
+  getRedHandler = async() => {
+    const goods = await getRedGoods();
+
+    this.setState({
+      goods,
+    });
+  };
```
aionova, line 35: You could make just 1 function and pass a callback into it.

aionova (changes requested):

aionova (approved):
Awesome! :rocket:

PR #223, opened 2021-01-25
first solution

Lara-l-art:
[DEMO LINK](https://Lara-l-art.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,75 @@
 import React from 'react';
-
 import './App.scss';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+const fiveGoods = [];
+const allResult = [];
+const redGoods = [];
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+function getArrayFromApi(func, array) {
+  return func().then(res => res.map(item => array.push(item)));
+}
+
+getArrayFromApi(getAll, allResult);
+getArrayFromApi(get5First, fiveGoods);
+getArrayFromApi(getRedGoods, redGoods);
+
+class App extends React.Component {
+  state = {
+    all: false,
+    five: false,
+    red: false,
+  }
+
+  render() {
+    return (
+      <>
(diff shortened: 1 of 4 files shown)


Serhii-Alieksieievych (approved):

aionova (changes requested):
Good start!
When you press 'Load 5 first goods', you should show only 5 first goods, no matter what button was pressed before that. The same should work for all the other buttons.
Screenshot 2021-01-26 at 09 51 11

src/App.js
```diff
@@ -1,13 +1,71 @@
 import React from 'react';
-
 import './App.scss';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { GoodsList } from './GoodsList';
+
+class App extends React.Component {
+  state = {
+    show: false,
```
aionova, line 8: You don't need show property here. 
When you press any button, you should immediately see the requested list.

aionova (changes requested):
Almost there)

src/App.js
```diff
@@ -1,13 +1,65 @@
...
-// import * as goodsAPI from './api/goods';
+        <button
+          type="button"
+          onClick={() => {
+            getAll()
+              .then((res) => {
+                this.setState(state => ({
+                  goods: res,
+                }));
+              });
```
aionova, line 28: Do all the logic outside return statement

src/App.js
```diff
@@ -1,13 +1,65 @@
...
+        </button>
+        <button
+          type="button"
+          onClick={() => {
+            get5First()
+              .then((res) => {
+                this.setState(state => ({
+                  goods: res,
+                }));
+              });
```
aionova, line 41: You can make 1 function outside return statement, that will accept a callback as a parameter and use it for 3 click handlers

aionova (changes requested):

src/App.js
```diff
@@ -1,13 +1,69 @@
...
 import './App.scss';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { GoodsList } from './GoodsList';
+
+class App extends React.Component {
+  state = {
+    goods: [],
+  }
+
+  setGoods = (res) => {
```
aionova, line 11:
```suggestion
  setGoods = (goods) => {
```

aionova (approved):
Great!

PR #224, opened 2021-01-25
Develop

HaidaiAndrii:
[DEMO LINK](https://HaidaiAndrii.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.js
+++ b/src/api/goods.js
@@ -6,10 +6,8 @@ export function getAll() {
     .then(response => response.json());
 }
 
-export const get5First = async() => {
-  [...await getAll()].slice(0, 5);
-};
+export const get5First = async() => [...await getAll()].slice(0, 5);
 
-export const getRedGoods = async() => {
-  [...await getAll()].filter(good => good.color === 'red');
-};
+export const getRedGoods = async() => (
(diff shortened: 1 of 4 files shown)


Lara-l-art (approved):
https://prntscr.com/xlwq5o
https://prntscr.com/xlwoyh

your list sorted but not as expected
try use "localeCompare" method

src/api/goods.js
```diff
@@ -6,6 +6,8 @@ export function getAll() {
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5First = async() => [...await getAll()].slice(0, 5);
 
-export const getRedGoods = () => {};
+export const getRedGoods = async() => (
+  [...await getAll()].filter(good => good.color === 'red')
```
aionova, line 12: Does filter mutate the original array?

aionova (changes requested):
As Larysa told you, you need to sort an array of goods and then slice first 5 of them

src/api/goods.js
HaidaiAndrii: nope)

aionova (approved):
Great!

PR #231, opened 2021-01-25
solution

dima-bond:
[DEMO LINK](https://dima-bond.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,12 +2,60 @@ import React from 'react';
 
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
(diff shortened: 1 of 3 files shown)


khrulenko (approved):

aionova (changes requested):
Nice start!

Screenshot 2021-01-26 at 10 05 50

First you need to sort items, and then to slice 5 of them.

src/App.js
```diff
@@ -2,12 +2,60 @@ import React from 'react';
...
+      .then((goods) => {
+        this.setState({ goods });
+      });
+  }
+
+  loadRedGoods = () => {
+    getRedGoods()
+      .then((goods) => {
+        this.setState({ goods });
+      });
```
aionova, line 31: You could make one function instead of these 3, that accepts callback as a parameter

aionova (approved):
Great!

PR #233, opened 2021-01-25
Solution

Roman-Matsuk:
[DEMO LINK](https://Roman-Matsuk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,12 +2,66 @@ import React from 'react';
 
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { GoodsList } from './api/components/GoodsList/GoodsList';
 // or
 // import * as goodsAPI from './api/goods';
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+class App extends React.Component {
+  state = {
+    goods: [],
+    isListVisible: false,
+  }
+
+  showAll = () => {
+    getAll()
+      .then((goods) => {
+        this.setState(prevState => ({
+          goods,
+          isListVisible: !prevState.isListVisible,
+        }));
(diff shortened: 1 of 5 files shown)


YuriiFitsyk (approved):
Nice!

src/App.js
```diff
@@ -2,12 +2,66 @@ import React from 'react';
...
+    goods: [],
+    isListVisible: false,
+  }
+
+  showAll = () => {
+    getAll()
+      .then((goods) => {
+        this.setState(prevState => ({
+          goods,
+          isListVisible: !prevState.isListVisible,
```
aionova, line 21: You don't need isListVisible property. 
You need to always render this.state.goods, and change it's value by callbacks. 
When I press showAll button, all the goods are rendered; show5First - first 5 are rendered. 
Now it doesn't work the way it should

aionova (changes requested):
Nice start!

src/api/goods.js
```diff
@@ -6,6 +6,14 @@ export function getAll() {
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5First = () => (
+  getAll()
+    .then(result => result
+      .sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5))
```
aionova, line 12:
```suggestion
      .sort((a, b) => a.name.localeCompare(b.name))
      .slice(0, 5))
```

src/App.js
```diff
@@ -2,12 +2,66 @@ import React from 'react';
...
+      .then((goods) => {
+        this.setState({
+          goods,
+          isListVisible: true,
+        });
+      });
+  }
+
+  render() {
+    const { isListVisible } = this.state;
```
aionova, line 47: these is no need in this property

aionova (approved):
Good job!

PR #236, opened 2021-01-25
Solution

ira-cherpak:
[DEMO LINK](https://ira-cherpak.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,52 @@
 import React from 'react';
+import { GoodsList } from './GoodsList';
 
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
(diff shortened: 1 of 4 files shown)


Serhii-Alieksieievych (approved):

src/App.js
```diff
@@ -1,13 +1,52 @@
...
+    const goods = await get5First();
+
+    this.setState({ goods });
+  };
+
+  getRedGoods = async() => {
+    const goods = await getRed();
+
+    this.setState({ goods });
+  };
```
aionova, line 29: You could make 1 function instead of these three, and pass callback as a parameter

aionova (approved):
Good job!

PR #241, opened 2021-01-26
added solution

MarchD:
[DEMO LINK](https://MarchD.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -8,22 +8,8 @@ class App extends React.Component {
     goods: [],
   }
 
-  selectAll = () => {
-    getAll()
-      .then((goods) => {
-        this.setState({ goods });
-      });
-  }
-
-  select5Goods = () => {
-    get5First()
-      .then((goods) => {
-        this.setState({ goods });
-      });
-  }
-
-  selectRedGoods = () => {
-    getRedGoods()
+  toSelect = (callback) => {
+    callback()
       .then((goods) => {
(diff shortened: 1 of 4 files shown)


Serhii-Alieksieievych (approved):

src/api/goods.js
```diff
@@ -6,6 +6,12 @@ export function getAll() {
...
-export const get5First = () => {};
+export const get5First = () => getAll()
+  .then(goods => (
+    goods.sort((a, b) => a.name.localeCompare(b.name)).splice(0, 5)
+  ));
 
-export const getRedGoods = () => {};
+export const getRedGoods = () => getAll()
+  .then(goods => (
+    goods.filter(good => good.color === 'red').splice(0, 5)
```
aionova, line 16: You don't need splice here. This function should get all the red goods, not only 5 of them

src/App.js
```diff
@@ -1,13 +1,56 @@
...
+      .then((goods) => {
+        this.setState({ goods });
+      });
+  }
+
+  selectRedGoods = () => {
+    getRedGoods()
+      .then((goods) => {
+        this.setState({ goods });
+      });
```
aionova, line 29: You could make just 1 function instead of these 3 and accept callback as a parameter

aionova (approved):
Good job!

PR #248, opened 2021-01-26
solution

dtsehelnyk:
[DEMO LINK](https://dtsehelnyk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,67 @@
 import React from 'react';
+import { GoodsList } from './components/GoodsList/GoodsList';
 
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRed } from './api/goods';
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+class App extends React.Component {
+  state = {
+    rightGoods: [],
+    selectedFilter: getAll,
+  }
+
+  componentDidUpdate() {
+    this.state.selectedFilter()
+      .then((goodsFromServer) => {
+        this.setState({
+          rightGoods: [...goodsFromServer],
+        });
+      });
+  }
+
+  selectFilter = (filter) => {
+    this.setState({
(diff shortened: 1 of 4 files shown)


troospieler (approved):
works good!

src/App.js
```diff
@@ -1,13 +1,67 @@
...
+          rightGoods: [...goodsFromServer],
+        });
+      });
+  }
+
+  selectFilter = (filter) => {
+    this.setState({
+      selectedFilter: filter,
+    });
+  }
```
aliceblunt8, line 27: no need to save it to state

src/App.js
```diff
@@ -1,13 +1,67 @@
...
+
+  componentDidUpdate() {
+    this.state.selectedFilter()
+      .then((goodsFromServer) => {
+        this.setState({
+          rightGoods: [...goodsFromServer],
+        });
+      });
+  }
```
aliceblunt8, line 21: don't do it in componentDidUpdate
let's create function getGoods, which will accept a callback (filter func) and call and then set to the state goods

aliceblunt8 (changes requested):

src/App.js
```diff
@@ -1,13 +1,60 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRed } from './api/goods';
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+class App extends React.Component {
+  state = {
+    rightGoods: [],
```
aliceblunt8, line 10:
```suggestion
    goods: [],
```

src/App.js
```diff
@@ -1,13 +1,60 @@
...
+  }
+
+  getGoods = (filter) => {
+    filter()
+      .then((goodsFromServer) => {
+        this.setState({
+          rightGoods: [...goodsFromServer],
```
aliceblunt8, line 17: no need to copy array

```suggestion
          rightGoods: goodsFromServer,
```

aliceblunt8 (approved):

PR #336, opened 2021-05-31
Dynamic list of goods

nazarmatsevych:
[DEMO LINK](https://nazarmatsevych.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,12 +2,64 @@ import React from 'react';
 
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { GoodsList } from './components/GoodsList';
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -2,12 +2,64 @@ import React from 'react';
...
+        this.setState({ goods });
+      });
+  }
+
+  showRedGoods = () => {
+    getRedGoods()
+      .then((goods) => {
+        this.setState({ goods });
+      });
+  }
```
rauzok, line 32: Your functions do the same.
I think you should create only one function and accept different callback for buttons.

rauzok (changes requested):
Nice work!

src/App.js
nazarmatsevych: agree, but now my code looks much better and clear

rauzok (approved):

aliceblunt8 (approved):

PR #346, opened 2021-06-01
solution

Namart:
[DEMO LINK](https://Namart.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/components/GoodsList.jsx
+++ b/src/components/GoodsList.jsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import propTypes from 'prop-types';
+
+export function GoodsList({ goods }) {
+  return (
+    <ul>
+      {goods.map(good => (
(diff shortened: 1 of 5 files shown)


src/components/GoodsList.jsx
```diff
@@ -0,0 +1,18 @@
...
+          {good.name}
+        </li>
+      ))}
+    </ul>
+  );
+}
+
+GoodsList.propTypes = {
+  goods: propTypes.arrayOf({}).isRequired,
+};
```
igor-stasiv, line 18: You need to set PropTypes for id, name, and color because you use it too

igor-stasiv (changes requested):
You forgot to set the color in styles for each good in the list. Also, add more propTypes

aliceblunt8 (approved):
GJ

PR #383, opened 2021-06-02
react_dynamic_list

Allrightman:
[DEMO LINK](https://Allrightman.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,68 @@
 import React from 'react';
 
 import './App.scss';
+import { GoodsList } from './components';
 
-// import { getAll, get5First, getRed } from './api/goods';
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -1,13 +1,68 @@
...
+class App extends React.Component {
+  state = {
+    goods: [],
+  }
+
+  handleGetGoods = () => (
+    getAll()
+      .then((goods) => {
+        this.setState({ goods });
+      })
```
VinnieJ-2k20, line 19: Potentially it can be taken out to some method, like `updateGoods` and used in all three handlers.

VinnieJ-2k20 (approved):
Good job 👍

PR #403, opened 2021-07-31
first-try

vitaliy-martynyuk:
[DEMO LINK](https://vitaliy-martynyuk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,12 +2,66 @@ import React from 'react';
 
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import GoodsList from './GoodsList';
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+class App extends React.PureComponent {
+  state = {
+    dataLoaded: false,
+    arrayOfGoods: [],
+  };
+
+  loadAllGoods = () => {
+    this.setState(getAll().then(array => this.setState({
+      arrayOfGoods: array,
+      dataLoaded: true,
+    })));
+  };
+
+  loadFirstFive = () => {
+    this.setState(get5First().then(array => this.setState({
+      arrayOfGoods: array,
+      dataLoaded: true,
+    })));
+  };
+
+  loadRed = () => {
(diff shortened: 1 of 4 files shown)


src/api/goods.js
```diff
@@ -6,6 +6,8 @@ export function getAll() {
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5First = () => getAll()
+  .then(result => result.slice(0, 5));
```
aleksandr-semerenko, line 10: Load 5 first goods shows 5 first goods after sorting them by name

aleksandr-semerenko (changes requested):

aleksandr-semerenko (approved):

src/api/goods.js
```diff
@@ -6,6 +6,9 @@ export function getAll() {
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5First = () => getAll()
+  .then(result => result.slice(0, 5)
+    .sort((a, b) => a.name.localeCompare(b.name)));
```
ArtemTopolskyi, line 11: is it necessary to sort them?

src/App.js
```diff
@@ -2,12 +2,66 @@ import React from 'react';
...
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import GoodsList from './GoodsList';
 
-const App = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+class App extends React.PureComponent {
+  state = {
+    dataLoaded: false,
```
ArtemTopolskyi, line 10: do you really need this variable? just pass variable arrayOfGoods to render method. It will be empty if data is not loaded yet

src/App.js
```diff
@@ -2,12 +2,66 @@ import React from 'react';
...
+    arrayOfGoods: [],
+  };
+
+  loadAllGoods = () => {
+    this.setState(getAll().then(array => this.setState({
+      arrayOfGoods: array,
+      dataLoaded: true,
+    })));
```
ArtemTopolskyi, line 18: are you sure, that it's good idea to call setState inside the setState? it looks weird. get rid of outer setState call

ArtemTopolskyi (changes requested):

PR #443, opened 2021-09-11
Develop

lera-yaskevych:
[DEMO LINK](https://lera-yaskevych.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,82 @@
 import React from 'react';
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,12 +1,119 @@
...
+            hasRequestError: true,
+          });
+
+          return null;
+        }
+
+      default:
+        return null;
+    }
+  };
```
ivanchenko-kirill, line 74: As for me, this a nice use of enum, quite understandable, but a bit huge 👍

ivanchenko-kirill (approved):
Nicely done 🚀

src/App.tsx
lera-yaskevych: It really is, you're right. Now that I've seen your approach I guess I'll refactor it, so that it'd take less space)
Thanks for that insight 🥇

danheim (approved):

PR #454, opened 2021-09-15
add list

evgen1sider:
[DEMO LINK](https://evgen1sider.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,69 @@
 import React from 'react';
 import './App.scss';
+import { AllGoodsButton } from './components/AllGoodsButton';
+import { GoodsList } from './components/GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { LoadFirst5Button } from './components/LoadFirst5Button';
+import { LoadRedGoodsButton } from './components/LoadRedGoodsButton';
 // or
 // import * as goodsAPI from './api/goods';
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+type State = {
+  goods: Good[] | null;
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    goods: null,
+  };
+
+  getAllGoods = () => {
+    getAll()
+      .then(goods => {
+        this.setState({
+          goods,
+        });
+      });
+  };
+
+  loadFirst5Goods = () => {
+    get5First()
+      .then(goods => {
+        this.setState({
+          goods,
+        });
+      });
+  };
+
+  loadRedGoods = () => {
+    getRedGoods()
+      .then(goods => {
+        this.setState({
+          goods,
+        });
+      });
+  };
+
+  render() {
+    return (
+      <div className="App container col-6">
+        <h1>Dynamic list of Goods</h1>
+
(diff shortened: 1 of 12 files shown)


src/App.tsx
```diff
@@ -1,12 +1,69 @@
...
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+type State = {
+  goods: Good[] | null;
+};
+
+class App extends React.Component<{}, State> {
+  state = {
```
Elena-brv, line 17:
```suggestion
  state: State = {
```

src/App.tsx
```diff
@@ -1,12 +1,69 @@
...
+
+        <div className="d-flex justify-content-between">
+          <AllGoodsButton getGoods={this.getAllGoods} />
+
+          <LoadFirst5Button getGoods={this.loadFirst5Goods} />
+
+          <LoadRedGoodsButton getGoods={this.loadRedGoods} />
+        </div>
+
+        {this.state.goods && (
```
Elena-brv, line 61: destrucruting

src/api/goods.ts
```diff
@@ -6,6 +6,14 @@ export function getAll(): Promise<Good[]> {
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5First = () => {
+  return getAll()
+    .then(goods => goods.sort((a, b) => a.name.localeCompare(b.name))
+      .filter(good => goods.indexOf(good) < 5));
```
Elena-brv, line 12:
```suggestion
  return getAll()
    .then(goods => goods
      .sort((a, b) => a.name.localeCompare(b.name))
      .filter(good => goods.indexOf(good) < 5)
    );
```
use slice instead of filter. Also filter get index as a second argument. So you don't need to use indexOf

src/components/GoodsList/GoodList.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+type Props = {
+  goods: Good[];
+};
+type State = {};
+
+export class GoodsList extends React.Component<Props, State> {
+  state: State = {};
```
Elena-brv, line 9: don't create an empty state. Use function component

src/App.tsx
```diff
@@ -1,12 +1,69 @@
...
+    return (
+      <div className="App container col-6">
+        <h1>Dynamic list of Goods</h1>
+
+        <div className="d-flex justify-content-between">
+          <AllGoodsButton getGoods={this.getAllGoods} />
+
+          <LoadFirst5Button getGoods={this.loadFirst5Goods} />
+
+          <LoadRedGoodsButton getGoods={this.loadRedGoods} />
```
Elena-brv, line 58: why do you need 3 different components for buttons? or make 1 component Button and pass needed props there. Or just place all buttons here

src/App.tsx
```diff
@@ -1,12 +1,69 @@
...
+  };
+
+  loadRedGoods = () => {
+    getRedGoods()
+      .then(goods => {
+        this.setState({
+          goods,
+        });
+      });
+  };
```
Elena-brv, line 46: the logic of all methods is similar. You can create one common method and use it by getting a needed callback as an argument

Elena-brv (changes requested):

Elena-brv (approved):

PR #459, opened 2021-09-15
React - Fetch - Dynamic list of goods

george-w-boo:
[DEMO LINK](https://george-w-boo.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,103 @@
 import React from 'react';
+import 'bulma';
+import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
 
 // import { getAll, get5First, getRed } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+import { getAll } from './api/goods';
+import { GoodsList } from './components/GoodsList';
+
+type Props = {};
+type State = {
+  goods: Good[] | [];
+};
+
+class App extends React.Component<Props, State> {
+  state: State = {
+    goods: [],
+  };
+
+  handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+
+    if (event.currentTarget.name === 'allGoods') {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,12 +1,96 @@
...
+
+    if (event.currentTarget.name === 'firstFive') {
+      getAll()
+        .then((goods: Good[]) => (
+          goods.filter((_, i) => i < 5)
+        ))
+        .then((x: Good[]) => {
+          this.setState({ goods: x });
+        });
+    }
```
DmytroMykoliv, line 41: By condition, you should sort users, after that return first five

src/api/goods.ts
```diff
@@ -6,6 +6,6 @@ export function getAll(): Promise<Good[]> {
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+// export const get5First = () => {};
```
DmytroMykoliv, line 9: By the condition, you should implement this function like below, and it will give you 5 users:
```suggestion
export const load5First = () => {
  return getAll()
    .then(users => users.sort((a, b) => a.name.localeCompare(b.name))
      .slice(0, 5));
};
```

src/api/goods.ts
```diff
@@ -6,6 +6,6 @@ export function getAll(): Promise<Good[]> {
...
 
-export const getRedGoods = () => {};
+// export const getRedGoods = () => {};
```
DmytroMykoliv, line 11: Same here

DmytroMykoliv (commented):
Working code, but I left some suggestions

src/App.tsx
george-w-boo: fixed, thanks

src/api/goods.ts
george-w-boo: fixed, thanks

src/api/goods.ts
george-w-boo: fixed, thanks

DmytroMykoliv (approved):
Good work )) 👍

danheim (approved):

PR #482, opened 2021-09-17
Develop

Vova-Stoliar:
[DEMO LINK](https://Vova-Stoliar.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,6 +6,7 @@ import { GoodsList } from './GoodsList/GoodsList';
 type State = {
   goods: Good[],
   showGoods: boolean,
+  showError: boolean,
 };
 
 export class App extends React.PureComponent<{}, State> {
@@ -16,15 +17,24 @@ export class App extends React.PureComponent<{}, State> {
       color: '',
     }],
     showGoods: false,
+    showError: false,
   };
 
   handleButton = async (goodsFilter: Promise<Good[]>) => {
-    const goods = await goodsFilter;
-
     this.setState({
-      goods,
-      showGoods: true,
+      showError: false,
     });
+
+    try {
+      const goods = await goodsFilter;
+
+      this.setState({
+        goods,
+        showGoods: true,
+      });
+    } catch (error) {
+      this.setState({ showError: true });
+    }
   };
 
   render() {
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -1,11 +1,26 @@
...
-export const getRedGoods = () => {};
+  return response.json();
+};
+
+export const get5First = () => {
+  const showFiveGoods = getAll().then(goods => goods
+    .sort((a, b) => a.name.localeCompare(b.name))
+    .slice(0, 5));
+
+  return showFiveGoods;
```
Elena-brv, line 19:
```suggestion
 return getAll()
   .then(goods => goods
     .sort((a, b) => a.name.localeCompare(b.name))
     .slice(0, 5)
   );
```

src/App.tsx
```diff
@@ -1,12 +1,75 @@
...
-);
+export class App extends React.PureComponent<{}, State> {
+  state:State = {
+    goods: [{
+      id: NaN,
+      name: '',
+      color: '',
+    }],
+    showGoods: false,
+    showError: false,
```
Elena-brv, line 20: is it not enough only one of these properties?

src/App.tsx
```diff
@@ -1,12 +1,75 @@
...
+      <>
+        <h1>Dynamic list of Goods</h1>
+        <div>
+          <button
+            type="button"
+            onClick={() => this.handleButton(getAll())}
+          >
+            Load All goods
+          </button>
+          {' '}
```
Elena-brv, line 51: why do you need this?

src/App.tsx
```diff
@@ -1,12 +1,75 @@
...
+          <button
+            type="button"
+            onClick={() => this.handleButton(getRedGoods())}
+          >
+            Load red goods
+          </button>
+        </div>
+        {this.state.showGoods && <GoodsList goods={this.state.goods} />}
+        {this.state.showError && (
```
Elena-brv, line 67: you can use only one of these properties and use ternary operator

Elena-brv (changes requested):

src/App.tsx
Vova-Stoliar: to have tiny space between buttons

Elena-brv: use css for this

Elena-brv (commented):
image

fix the API link

PR #571, opened 2021-12-17
Develop

Val-space:
[DEMO LINK](https://Val-space.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,79 @@
+/* eslint-disable no-console */
 import React from 'react';
 import './App.scss';
+import 'bulma/css/bulma.min.css';
+import { GoodsList } from './components/GoodsList/GoodsList';
+import { LoadingError } from './components/LoadingError/LoadingError';
+import { Loader } from './components/Loader/Loader';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+type State = {
+  goods: Good[] | [];
+  isLoading: boolean;
+  hasLoadingError: boolean;
+};
+
+class App extends React.Component<{}, State> {
+  state:State = {
+    goods: [],
+    isLoading: false,
+    hasLoadingError: false,
+  };
+
+  loadData = async (getData: () => Promise<Good[]>) => {
+    this.setState({
+      isLoading: true,
+    });
+
+    try {
+      const goods = await getData();
+
+      this.setState({
+        goods,
+        isLoading: false,
+      });
+    } catch {
+      this.setState({
+        hasLoadingError: true,
+      });
+    }
+  };
+
+  render() {
+    const { goods, hasLoadingError, isLoading } = this.state;
+
+    return (
+      <div className="container">
+        <h1 className="title">GOODS</h1>
+        <button
+          type="button"
+          className="button is-link m-5"
+          onClick={() => this.loadData(getAll)}
+        >
+          Load All goods
+        </button>
+
+        <button
+          type="button"
+          className="button is-success m-5"
+          onClick={() => this.loadData(get5First)}
+        >
+          get5First
+        </button>
+
+        <button
+          type="button"
+          className="button is-danger m-5"
+          onClick={() => this.loadData(getRedGoods)}
+        >
+          Load red goods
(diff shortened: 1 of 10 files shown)


src/api/goods.ts
```diff
@@ -1,11 +1,29 @@
...
+
   return fetch(API_URL)
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5First = async () => {
+  const [allGoods] = await Promise.all([
+    getAll(),
+  ]);
```
dmtrhrytsak, line 18: Hmm, I don't think it `Promise.all()` does anything useful in this case😅 Promises should be combined in one `Promise.all()` for it to work (Send multiple requests at the same time)

src/App.tsx
```diff
@@ -1,12 +1,79 @@
...
+      const goods = await getData();
+
+      this.setState({
+        goods,
+        isLoading: false,
+      });
+    } catch {
+      this.setState({
+        hasLoadingError: true,
+      });
```
dmtrhrytsak, line 39: If the error occurs, `isLoading` won't become false, and you know what happens next)
Set `isLoading` to false in the `final` clause, this way you can even omit it in the `try` clause

dmtrhrytsak (approved):
Goods are freaking good!✨

src/api/goods.ts
Val-space: i agree, just wanted to try it but then realized it doesn't do its job here

src/App.tsx
```diff
@@ -1,12 +1,80 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+type State = {
+  goods: Good[] | [];
```
illiarerun, line 12: No need for an empty arrray type
```suggestion
  goods: Good[];
```

src/App.tsx
```diff
@@ -1,12 +1,80 @@
...
+  isLoading: boolean;
+  hasLoadingError: boolean;
+};
+
+class App extends React.Component<{}, State> {
+  state:State = {
```
illiarerun, line 18:
```suggestion
  state: State = {
```

src/api/goods.ts
```diff
@@ -1,11 +1,25 @@
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll(): Promise<Good[]> {
+function wait(delay: number) {
+  return new Promise(resolve => setTimeout(resolve, delay));
+}
+
+export async function getAll(): Promise<Good[]> {
+  await wait(2000);
```
illiarerun, line 9: 2 seconds is waay tooo much, you can literally remove it, or make it as small as possible

src/components/Loader/loader.css
```diff
@@ -0,0 +1,59 @@
+.bar {
+  width: 10px;
+  height: 70px;
+  background: #fff;
+  display: inline-block;
+  transform-origin: bottom center;
+  border-top-right-radius: 20px;
+  border-top-left-radius: 20px;
+  /*   box-shadow:5px 10px 20px inset rgba(255,23,25.2); */
```
illiarerun, line 9: Commented code 👿

src/components/Loader/loader.css
```diff
@@ -0,0 +1,59 @@
...
+  animation-delay: 0.6s;
+}
+
+.bar7 {
+  animation-delay: 0.7s;
+}
+
+.bar8 {
+  animation-delay: 0.8s;
+}
```
illiarerun, line 43: You can use sccs loop here

src/App.tsx
```diff
@@ -1,12 +1,80 @@
...
+
+        <button
+          type="button"
+          className="button is-danger m-5"
+          onClick={() => this.loadData(getRedGoods)}
+        >
+          Load red goods
+        </button>
+        {hasLoadingError && <LoadingError /> }
+        {isLoading ? <Loader /> : <GoodsList goods={goods} />}
```
illiarerun, line 74: It would be rendered even if error occurred, right? You should fix it

illiarerun (changes requested):
GJ 🔥 Let's work on fixes now:

src/App.tsx
Val-space: right..right..

PR #582, opened 2021-12-20
Solution

vlad-tuzenko:
[DEMO LINK](https://vlad-tuzenko.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,75 @@
+/* eslint-disable no-console */
 import React from 'react';
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { GoodsList } from './components/GoodsList';
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+type State = {
+  goods: Good[];
+  isLoading: boolean;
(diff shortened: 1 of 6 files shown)


olya-shyian (approved):

src/App.tsx
```diff
@@ -1,12 +1,75 @@
...
+          <button type="button" onClick={() => this.loadGoods(getAll)}>
+            Load All goods
+          </button>
+          <button type="button" onClick={() => this.loadGoods(get5First)}>
+            Load 5 first goods
+          </button>
+          <button type="button" onClick={() => this.loadGoods(getRedGoods)}>
+            Load red goods
+          </button>
+          {console.log(this.state.goods)}
```
AlinaMatuschak, line 57: delete console

src/App.tsx
```diff
@@ -1,12 +1,75 @@
...
+        </div>
+        <div className="App__content">
+          {isLoading && (
+            <div className="App__content-loading">Loading... Please, wait</div>
+          )}
+          {goods.length !== 0 && (
```
AlinaMatuschak, line 63:
```suggestion
          {goods.length && (
```

AlinaMatuschak (approved):

src/App.tsx
vlad-tuzenko: image
In this case i see 0 while loading the list, so I replace that with the following: ``!!goods.length``

PR #621, opened 2022-01-31
My solution

SavelovSerhii:
https://savelovserhii.github.io/react_dynamic-list-of-goods/


src/App.tsx
```diff
@@ -1,12 +1,60 @@
+/* eslint-disable react/no-unused-state */
```
Olexandr6, line 1: Don't disable linter

src/App.tsx
```diff
@@ -1,12 +1,60 @@
...
+    return (
+      <div className="App">
+        <h1>Dynamic list of Goods</h1>
+        <button
+          type="button"
+          onClick={async () => {
+            this.setState({
+              goodsFromServer: await getAll(),
+            });
+          }}
```
Olexandr6, line 27: Create universal handler, and pass there a callback, for example: getGoods = `async (callback: () => Promise<Good[]>) {...}`

Olexandr6 (changes requested):
Good job, but some things should be optimized

Olexandr6 (approved):

PR #654, opened 2022-02-03
solution "Dynamic list of goods"

IrynaZolko:
[DEMO LINK](https://IrynaZolko.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,61 @@
 import React from 'react';
+import 'bulma';
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll } from './api/goods';
+import { GoodsButtons } from './components/GoodsButtons';
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+type Props = {};
+type State = {
+  goods: Good[],
+  error: boolean,
+};
 
-export default App;
+export class App extends React.Component<Props, State> {
+  state: State = {
+    goods: [],
+    error: false,
+  };
+
+  async componentDidMount() {
+    try {
+      const goods = await getAll();
+
+      this.setState({
+        goods: [...goods],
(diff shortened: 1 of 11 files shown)


src/App.tsx
```diff
@@ -1,12 +1,61 @@
...
+    this.setState({
+      goods: [...goods],
+    });
+  };
+
+  render() {
+    const { goods, error } = this.state;
+
+    // eslint-disable-next-line no-console
+    console.log('goods', goods);
```
Olexandr6, line 46: Don't leave comments in the production code

src/App.tsx
```diff
@@ -1,12 +1,61 @@
...
+
+    if (error) {
+      return (
+        <div>error</div>
+      );
+    }
```
Olexandr6, line 52: Use conditional rendering instead

src/api/goods.ts
```diff
@@ -1,11 +1,31 @@
...
-  return fetch(API_URL)
-    .then(response => response.json());
+export async function getAll(): Promise<Good[]> {
+  try {
+    const response = await fetch(API_URL);
+
+    return await response.json();
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.log('error', error);
```
Olexandr6, line 11: Don't leave comments

src/App.tsx
```diff
@@ -1,12 +1,61 @@
...
+        goods: [...goods],
+      });
+    } catch (e) {
+      this.setState({
+        error: true,
+      });
+    }
+  }
+
+  loadGoods = async (getGoods: () => Promise<Good[]>) => {
```
Olexandr6, line 34: 🔥

src/App.tsx
Olexandr6: Avoid multiple `return` statements in the `render()` method

Olexandr6 (changes requested):
Well done 👍  Check comments, some small things should be fixed

PR #660, opened 2022-02-21
task complete

voronvasia:
https://voronvasia.github.io/react_dynamic-list-of-goods/


src/App.tsx
```diff
@@ -1,12 +1,48 @@
...
+    const goodsFromServer = await goodsAPI.get5First();
+
+    setGoods(goodsFromServer);
+  };
+
+  const getRedGoods = async () => {
+    const goodsFromServer = await goodsAPI.getRedGoods();
+
+    setGoods(goodsFromServer);
+  };
```
aliceblunt8, line 26: these functions look almost the same, let's create one function that receives callback, call it and write the result to the state

aliceblunt8 (changes requested):

PR #667, opened 2022-03-17
Develop

mariiaStorozhyk:
[DEMO LINK](https://mariiaStorozhyk.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,12 +1,46 @@
...
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+
+  const getAllGoods = (getFunction: () => Promise<Good[]>) => {
+    getFunction()
+      .then(goodsFromServer => {
```
Elena-brv, line 12: try to use async/await, it's more common in real projects

src/GoodsList.tsx
```diff
@@ -0,0 +1,22 @@
+import { Good } from './Good';
+
+type Props = {
+  goods: Good[]
+};
+
+export const GoodsList: React.FC<Props> = ({ goods }) => {
+  return (
```
Elena-brv, line 8:
```suggestion
export const GoodsList: React.FC<Props> = ({ goods }) => (
```

src/GoodsList.tsx
```diff
@@ -0,0 +1,22 @@
...
+    <div>
```
Elena-brv, line 9: is this div necessary here?

Elena-brv (approved):

src/GoodsList.tsx
mariiaStorozhyk: i don't think so)

PR #699, opened 2022-04-11
solution

VladBrynzia:
[DEMO LINK](https://vladbrynzia.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,12 +1,47 @@
...
+        setGoods(good);
+      });
+  };
+
+  const getRedGoods = () => {
+    getRed()
+      .then(good => {
+        setGoods(good);
+      });
+  };
```
mariiishka, line 28: code duplication, create one function that takes a callback

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,17 @@
...
+  goods: Good[],
+};
+
+export const GoodsList: FC<Props> = ({ goods }) => {
+  return (
+    <ul>
+      {goods.map(good => (
+        <li key={good.id}>
+          {good.name}
+        </li>
```
mariiishka, line 13: it will be good if you add styles (color for example)

mariiishka (changes requested):

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,17 @@
+import { FC } from 'react';
+
+type Props = {
+  goods: Good[],
+};
+
+export const GoodsList: FC<Props> = ({ goods }) => {
```
solaryasha, line 7: use React.memo

PR #712, opened 2022-04-19
v1

KristinaTar:


src/App.tsx
```diff
@@ -1,12 +1,52 @@
...
+          setGoods(await getAll());
+        }}
+      >
+        Load all goods
+      </button>
+      <button
+        type="button"
+        onClick={async () => {
+          setGoods(await get5First());
+        }}
```
Bohdan-Maksymiuk, line 32: you may create separate universal method which accepts callback from API

Bohdan-Maksymiuk (changes requested):
Good job but you can do some optimization

Bohdan-Maksymiuk (approved):

PR #769, opened 2022-06-10
Solution

DariaVeretyak:
https://dariaveretyak.github.io/react_dynamic-list-of-goods/


Bohdan-Maksymiuk (changes requested):
https://user-images.githubusercontent.com/94289038/173347749-2bc3825f-ccf4-49f6-840e-e84af47bad42.mov

All click on buttons should change the list, not hiding it on the next click

PR #771, opened 2022-06-13
SolutionI

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -1,11 +1,23 @@
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll(): Promise<Good[]> {
-  return fetch(API_URL)
-    .then(response => response.json());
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -1,11 +1,23 @@
...
+export async function getAll(): Promise<Good[]> {
+  const response = await fetch(API_URL);
+  const res = await response.json();
+
+  return res;
 }
 
-export const get5First = () => {};
+export const get5First = async () => {
+  const response = await fetch(API_URL);
```
KostiaO, line 12: you can recall getAll()

KostiaO (changes requested):

src/api/goods.ts
```diff
@@ -1,11 +1,25 @@
...
+  const res = await response.json();
```
Anna-Tkachenko, line 13: Agree, will be better to call getAll here (Don't repeat yourself)

Anna-Tkachenko (approved):

PR #778, opened 2022-06-14
Develop

vlad-demchuk:
[DEMO LINK](https://vlad-demchuk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,12 +1,67 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRed } from './api/goods';
+import { GoodsList } from './components/GoodsList';
+import { Good } from './react-app-env';
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+
+  const showAll = () => {
+    getAll()
+      .then((goodsFromServer) => {
+        setGoods(goodsFromServer);
+      });
+  };
+
+  const showFive = () => {
+    get5First()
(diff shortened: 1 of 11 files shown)


src/api/goods.ts
```diff
@@ -1,11 +1,18 @@
...
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5First = () => {
+  return getAll()
+    .then(goods => goods.slice(0, 5));
+};
```
RomekIvantsiv, line 13: shows 5 first `goods` after **sorting** them by name

RomekIvantsiv (changes requested):
Good job, add just small fixes)

src/api/goods.ts
vlad-demchuk: really, thanks

src/App.tsx
```diff
@@ -1,12 +1,58 @@
...
+import { Good } from './react-app-env';
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+
+  const showAll = async () => {
+    setGoods(await getAll());
```
Anna-Tkachenko, line 12: Will be better to save the result of async function in some variable and then use this variable for sending data in setGoods method

src/api/goods.ts
```diff
@@ -1,11 +1,24 @@
...
-export function getAll(): Promise<Good[]> {
-  return fetch(API_URL)
-    .then(response => response.json());
+export async function getAll(): Promise<Good[]> {
+  const response = await fetch(API_URL);
+
+  return response.ok
+  // is it okay?
+    ? response.json()
+    : Promise.reject(new Error(`${response.status}: ${response.statusText}`));
```
Anna-Tkachenko, line 11: Will be better to use try / catch with async / await

Anna-Tkachenko (approved):

PR #781, opened 2022-06-14
solution

alexandra-protyanova:


src/App.tsx
```diff
@@ -1,12 +1,74 @@
...
+    setVisibleGoods(goods);
+  };
+
+  const firstFiveCashedGoods = () => {
+    if (goods.length === 0) {
+      goodsFromServer();
+    }
+
+    setVisibleGoods([...goods
+      .sort((a, b) => a.name.localeCompare(b.name))].slice(0, 6));
```
vlasiuk-anatolii, line 32: ].slice(0, 5));

vlasiuk-anatolii (approved):
Great Job!

src/App.tsx
alexandra-protyanova: When I want to take first five elements, I should do it like `slice(0, 6)` , slice take all elements exsepts index 6)

src/api/goods.ts
```diff
@@ -1,11 +1,8 @@
...
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
     .then(response => response.json());
 }
-
-export const get5First = () => {};
-
-export const getRedGoods = () => {};
```
Anna-Tkachenko, line 11: You shouldn't delete this functions
You should use them in your code
get5First function should return an array with 5 elements
getRedGoods function should return an array with red elements

Anna-Tkachenko (changes requested):

src/components/GoodsList.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Good } from '../react-app-env';
+import './GoodsList.scss';
+
+type Props = {
+  goods: Good[],
+};
+
+export const GoodsList: React.FC<Props> = React.memo(
+  ({ goods }) => {
+    return (
```
Denchepornyuk, line 11: remove return from GoodsList component

Denchepornyuk (approved):

PR #783, opened 2022-06-15
Develop

AnastasiyaSopotnytska:
[DEMO LINK](https://AnastasiyaSopotnytska.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 
 import { get5First, getAll, getRedGoods } from './api/goods';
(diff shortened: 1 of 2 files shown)


vlasiuk-anatolii (approved):

src/App.tsx
```diff
@@ -1,12 +1,84 @@
...
+import { GoodList } from './components/GoodsList';
 
-const App: React.FC = () => (
-  <h1>Dynamic list of Goods</h1>
-);
+const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [hasAll, setHasAll] = useState(false);
+  const [hasFive, setHasFive] = useState(false);
+  const [hasRed, setHasRed] = useState(false);
```
Anna-Tkachenko, line 12: This variables are redundant

src/App.tsx
```diff
@@ -1,12 +1,84 @@
...
+    if (hasFive) {
+      get5First()
+        .then(shownGoods => {
+          const shown5Goods = shownGoods
+            .sort((a: Good, b: Good | any) => a.name.localeCompare(b.name))
+            .splice(0, 5);
+
+          setGoods(shown5Goods);
+        });
+    }
```
Anna-Tkachenko, line 41: Set all goods to the state by default

src/App.tsx
```diff
@@ -1,12 +1,84 @@
...
+
+  return (
+    <>
+      <button
+        type="button"
+        onClick={() => {
+          setHasFive(false);
+          setHasRed(false);
+          setHasAll(true);
+        }}
```
Anna-Tkachenko, line 52: Call here only one needed function for every button

src/api/goods.ts
```diff
@@ -1,11 +1,15 @@
...
-export function getAll(): Promise<Good[]> {
+export const getAll = () => {
   return fetch(API_URL)
     .then(response => response.json());
-}
+};
 
-export const get5First = () => {};
+export const get5First = () => {
+  return getAll();
```
Anna-Tkachenko, line 10: You should return array with only 5 elements from this function

src/api/goods.ts
```diff
@@ -1,11 +1,15 @@
...
+};
 
-export const getRedGoods = () => {};
+export const getRedGoods = () => {
+  return getAll();
```
Anna-Tkachenko, line 14: You should return array only with red goods from this function

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #790, opened 2022-06-15
Solution

v-vytrykush:
[DEMO](https://v-vytrykush.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -1,11 +1,21 @@
+import { Good } from '../react-app-env';
+
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll(): Promise<Good[]> {
+export function getAllGoods(): Promise<Good[]> {
   return fetch(API_URL)
     .then(response => response.json());
 }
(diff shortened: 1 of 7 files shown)


src/api/goods.ts
```diff
@@ -1,11 +1,21 @@
...
   return fetch(API_URL)
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5Goods = (num:number): Promise<Good[]> => {
+  return getAllGoods()
+    // eslint-disable-next-line max-len
+    .then(result => result.sort((firstGood, secondGood) => firstGood.name.localeCompare(secondGood.name)))
+    .then(result => result.slice(0, num));
```
ZUBOV-ILLIA, line 15: I think that if you will do first slice and after cort it can be much faster operation !? @Anna-Tkachenko

ZUBOV-ILLIA (approved):
Good Job!

src/api/goods.ts
Anna-Tkachenko: Yes, will be better to do it for code optimization

Anna-Tkachenko (approved):

PR #792, opened 2022-06-15
Task Done v1.0

khozhainov-aleksandr:
[DEMO LINK](https://khozhainov-aleksandr.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,12 +1,65 @@
...
-// import * as goodsAPI from './api/goods';
+const App: React.FC = () => {
+  const [allGoods, setAllGoods] = useState<Good[]>([]);
+
+  function getAllGoods() {
+    getAll()
+      .then((goods) => setAllGoods(goods));
+  }
+
+  function get5First() {
```
iwanttosmile4u, line 16: I think, that this function (with sorting and slicing) should be in goods.ts file, and here you should only change the state. But I'm not sure if it's a mistake if that logic is in the component App.

src/App.tsx
```diff
@@ -1,12 +1,65 @@
...
+    getAll()
+      .then((goods) => (
+        setAllGoods(goods
+          .sort((a, b) => a.name.localeCompare(b.name))
+          .slice(0, 5))
+      ));
+  }
+
+  function getRedGoods() {
```
iwanttosmile4u, line 25: The same recommendation in this case too.

iwanttosmile4u (commented):
Good job 👍 Please, look at my comment)

src/App.tsx
```diff
@@ -1,12 +1,53 @@
...
+
+const App: React.FC = () => {
+  const [allGoods, setAllGoods] = useState<Good[]>([]);
+
+  const getGoods = (callBack: CallBack) => {
+    callBack()
+      .then(goodsFromServer => {
+        setAllGoods(goodsFromServer);
+      });
+  };
```
Anna-Tkachenko, line 18: 🔥

Anna-Tkachenko (approved):
Good job!

PR #798, opened 2022-06-19
First commit

igoree88:
[DEMO LINK](https://igoree88.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,12 +1,68 @@
...
+  return (
+    <div className="app box">
+      <h1
+        className="title is-3"
+      >
+        Dynamic list of Goods
+      </h1>
+
+      {goods && (
+        <GoodsList goods={goods} />)}
```
enrikadej, line 20: it would be nice to relocate this code after buttons, (the list will be changed and buttons will stay at their initial place)

src/api/goods.ts
```diff
@@ -6,6 +6,16 @@ export function getAll(): Promise<Good[]> {
     .then(response => response.json());
 }
 
-export const get5First = () => {};
+export const get5First = ():Promise<Good[]> => {
+  return fetch(API_URL)
+    .then(response => response.json())
+    .then(goodsFromServer => goodsFromServer
+      .filter((good: { id: number; }) => good.id <= 5));
```
enrikadej, line 13: Unless I'm wrong, in function 'get5first' goods, should be sorted by name (first point of the checklist)

src/App.tsx
```diff
@@ -1,12 +1,68 @@
...
+
+      <div className="app__buttons">
+        <button
+          className="button is-primary"
+          type="button"
+          onClick={() => {
+            getAll().then(goodsFromServer => {
+              setGoods(goodsFromServer);
+            });
```
enrikadej, line 29: create separate function for each button

enrikadej (changes requested):
Good Job!

PR #807, opened 2022-06-28
solution

LukaszGryziewicz:
No styles for now, might have some fun with it later
 [DEMO LINK](https://LukaszGryziewicz.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -1,11 +1,24 @@
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll(): Promise<Good[]> {
-  return fetch(API_URL)
-    .then(response => response.json());
-}
+export const getAll = async (): Promise<Good[]> => {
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -1,11 +1,24 @@
...
+    throw new Error(`${error.status} - ${error.statusText}`);
+  }
+};
 
-export const getRedGoods = () => {};
+export const get5First = async () => {
+  const goods = await getAll();
+
+  return goods.slice(0, 5);
+};
```
amitpatiljc, line 18: Actually it not only first 5 its 
Load 5 first goods shows 5 first goods after sorting them by name

amitpatiljc (changes requested):
Great Job! Just one change is required.

src/api/goods.ts
LukaszGryziewicz: I missed that, thank you!

amitpatiljc (approved):
Good Job!

PR #817, opened 2022-06-30
dynamic list

Sardorbek1122:
[DEMO LINK](https://Sardorbek1122.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/components/GoodList.tsx
+++ b/src/components/GoodList.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+
+interface Props {
+  goods: Good[];
+}
+
+export const GoodsList: React.FC<Props> = React.memo(
+  ({ goods }) => {
+    return (
+      <ul>
+        {goods.map(good => (
+          <li
+            key={good.id}
+            style={{ color: good.color }}
+          >
+            {good.name}
+          </li>
+        ))}
+      </ul>
+    );
+  },
+);


59MrRobot (approved):
👍

src/App.tsx
```diff
@@ -1,12 +1,65 @@
...
+    const loadedGoods = await get5First();
+
+    setGoods(loadedGoods);
+  };
+
+  const loadRedGoods = async () => {
+    const loadedGoods = await getRed();
+
+    setGoods(loadedGoods);
+  };
```
VinnieJ-2k20, line 26: useCallback

src/api/goods.ts
```diff
@@ -1,11 +1,27 @@
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll(): Promise<Good[]> {
+export const getAll = (): Promise<Good[]> => {
```
VinnieJ-2k20, line 4: If you use arrow functions - it's better to type variable rather than function parts.
Create separate type for getAll, get5First and getRed
`export const getAll: CallbackType = () => {...}`

VinnieJ-2k20 (changes requested):
Almost done, let's optimise your code a little bit.

meforoma (changes requested):
hmm, it looks fine and deployment works, but I can't find `GoodsList` file
`{ GoodsList } from './components/GoodList';`
actually even the entire `./components` directory ¯\_(ツ)_/¯
commit all the files please.

meforoma (approved):

PR #840, opened 2022-07-25
solution

AKomashchuk:
https://akomashchuk.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,65 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
-  </div>
-);
+    setGoods(data);
+  };
+
+  const getRedColorGoods = async () => {
+    const data = await getRedGoods();
+
+    setGoods(data);
+  };
```
Olexandr6, line 27: Let's create a `loadGoods` function and pass then a callback (getAll, get5First, getRedGoods), instead of creating 3 similar functions

Olexandr6:
```
 loadGoods = async (callback: () => Promise<Good[]>) => {
		....
		}
```

Olexandr6 (changes requested):
Almost done🔥 Let's improve your solution a bit👇

Olexandr6 (approved):
Gj🔥

PR #866, opened 2022-07-26
finished with dynamic goods without styles

omelika:
[DEMO LINK](https://omelika.github.io/react_dynamic-list-of-goods/)


src/api/goods.ts
```diff
@@ -10,10 +10,12 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods.slice(0, 5)
+      .sort((a, b) => a.name.length - b.name.length)); // sort and get the first 5
```
Olexandr6, line 14: You should sort at first and then slice sorted array to get right results

Olexandr6 (changes requested):

Olexandr6 (approved):

PR #880, opened 2022-08-03
Solution

mikhail-shchuka:
https://mikhail-shchuka.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,58 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,58 @@
...
-      Load all goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
+          getAll().then((allGoods) => {
+            setGoods(allGoods);
+          });
+        }}
```
Anna-Tkachenko, line 26: You can also create one general function, which can accept goal function as an argument (getAll, get5First, getRedGoods), call it and set the result in state:
generalFunction().then(...)

It will be better than repeat one code 3 times, as you did

Anna-Tkachenko (approved):

PR #883, opened 2022-08-05
solution

nastia-sydorchuk:
[DEMO LINK](https://nastia-sydorchuk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,12 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,12 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
```
alexandra-protyanova, line 11: let us think about situation where server will response with error

alexandra-protyanova (changes requested):

ArturErmolenko (approved):
Well done 👍

PR #889, opened 2022-08-11
solution

DarkMistyRoom:
[DEMO LINK](https://DarkMistyRoom.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,48 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,48 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
```
ArturErmolenko, line 6: Let us think about situation when error comes from server

ArturErmolenko (changes requested):

valerii-derkach (approved):

PR #908, opened 2022-08-20
Solution

Lentyaeev:
[DEMO LINK](https://lentyaeev.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,53 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRed } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  return (
+    <div className="App panel">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        className="button"
+        data-cy="all-button"
+        onClick={() => {
+          getAll().then(response => setGoods(response));
```
blcts, line 22: .then(response => setGoods(response)); вынеси отдельно, чтоб не повторяться

src/api/goods.ts
```diff
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
...
+    .then(goods => goods.sort((a, b) => a.name.localeCompare(b.name)))
+    .then(goods => goods.slice(0, 5));
 };
 
-export const getRedGoods = () => {
+export const getRed = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods)
+    .then(goods => goods.filter(good => good.color === 'red'));
```
blcts, line 21: ты можешь это записывать так 
return getAll()

blcts (changes requested):

src/api/goods.ts
```diff
@@ -10,10 +10,11 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods)
```
ivan-kadykalo, line 13: Why do you need this line? It makes nothing
```suggestion
```

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):
🚀
Would be good to set the loader during downloading data, because firstly I thought that buttons don't work.

PR #933, opened 2022-08-22
done

butenko108:
[DEMO LINK](https://butenko108.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,65 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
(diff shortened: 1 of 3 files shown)


HennadiiKupr:
Nice

src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+  }, [goodsRequest]);
+
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
+
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => setGoodsRequest('all')}
```
ivan-kadykalo, line 41: I'm not sure that it is a good idea:
You are setting "goodsRequest", your page rendering, and then getting data from the server, set it and make a second render with the data. So you have 2 renders, and get data only after the second render.

Will be better if you will call the method onClick which gets goods and set it immediately, so you will get all the data only in one render.

ivan-kadykalo (changes requested):

src/App.tsx
butenko108: To be honest, I don't understand what you mean.

ivan-kadykalo (approved):
🔥

PR #935, opened 2022-08-23
solution

bogdana18:
[DEMO LINK](https://bogdana18.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,11 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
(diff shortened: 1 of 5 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,11 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
```
alexandra-protyanova, line 12: let us think about situation when we won't get data from server or get error

alexandra-protyanova (changes requested):
Add catching errors

ArturErmolenko (approved):

PR #947, opened 2022-09-07
develop

YaroslavKolbaiev:
[DEMO LINK](https://YaroslavKolbaiev.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,71 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import 'bulma';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goodsFromApi, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const fetchAllGoods = async () => {
+    const goods = await getAll();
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,71 @@
...
+          onClick={fetch5First}
+        >
+          Load 5 first goods
+        </button>
+
+        <button
+          className="button is-primary is-light"
+          type="button"
+          data-cy="red-button"
+          onClick={fetchRedGoods}
```
alexandra-protyanova, line 61: set red color to all red goods

alexandra-protyanova (changes requested):

src/GoodsList.tsx
```diff
@@ -1,14 +1,22 @@
...
 
-export const GoodsList: React.FC<Props> = ({ goods }) => (
+export const GoodsList: React.FC<Props> = ({ goods, red }) => (
   <ul className="has-text-centered">
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li
+        key={good.id}
+        data-cy="good"
+        className={classNames('has-text-black', {
```
TarasHoliuk, line 16: Unfortunately, it's still not red in the demo:

image

And that's not only about red goods:
image

Nice Bulma styles! but with those colors would be easier to just use `color` property from `good` object

src/GoodsList.tsx
```diff
@@ -1,14 +1,22 @@
+import classNames from 'classnames';
```
TarasHoliuk, line 1: one of the reasons why the red color doesn't work could be the absence of `classnames` in `package.json` dependencies. It adds automatically when you run `npm install classnames`

TarasHoliuk (changes requested):
Good job! but you have to fix colors

PR #953, opened 2022-09-15
solution

Kulchevych:
[DEMO LINK](https://Kulchevych.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,43 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  return (
+    <div className="App">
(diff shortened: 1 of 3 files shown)


src/GoodsList.tsx
```diff
@@ -7,10 +7,14 @@ type Props = {
 
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
-        {good.name}
-      </li>
-    ))}
+    {goods.map(good => {
```
Githi54, line 10:
```suggestion
    {goods.map(({ id, name, color }) => {
```
:)

src/GoodsList.tsx
```diff
@@ -7,10 +7,14 @@ type Props = {
...
+      const { id, name, color } = good;
```
Githi54, line 11:
```suggestion
 
```

src/App.tsx
```diff
@@ -1,27 +1,43 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={async () => setGoods(await getAll())}
```
Githi54, line 18: I think you can export a function and pass a callback to it

Githi54 (approved):
Great!🚀

src/App.tsx
dellachaise, line 18: Create a separate method that will accept as a parameter loader function and set it in state. And use it in each `Onclick`.

dellachaise (changes requested):
Create a separate method that will accept as a parameter loader function and set it in state. And use it in each `Onclick`.

dellachaise (approved):
Well done 👍

PR #967, opened 2022-09-26
dynamic list of goods solution

zinchenko-anastasiia:
[DEMO LINK](https://zinchenko-anastasiia.github.io/react_dynamic-list-of-goods/)


oksana-logos-frontend (approved):
Good job))

src/App.tsx
```diff
@@ -1,27 +1,47 @@
...
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => getAll()
+          .then(goodsFromServer => setGoods(goodsFromServer))}
+      >
```
dellachaise, line 22: Create separate methods instead of creating inline event handlers.

And you write kind of the same code 3 times. Better create a helper function that would receive `callback` as a parameter and `setGoods` for needed `goods`.
And then call it for each `onClick` with the needed callback function as a parameter.

src/GoodsList.tsx
```diff
@@ -8,7 +8,11 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li
```
dellachaise, line 11: Use destructuring in line 10.

src/api/goods.ts
```diff
@@ -10,10 +10,12 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((a, b) => a.name.localeCompare(b.name))
```
dellachaise, line 14: `a and b` is not good and meaningful names. Rename.

dellachaise (changes requested):
Good job 👍 

Just add the necessary changes according to comments.

Optional: The react tasks you're currently working on are great to try out some css framework and see how it works.
for example, to use [bulma](https://bulma.io/), just run the command
`npm install bulma`
import bulma in the main index.js file `index.tsx`
`import 'bulma/css/bulma.min.css';`
and all :heavy_check_mark:
you can apply the necessary classes to the components and see how it looks :tada:

dellachaise (approved):
Well done 👍

PR #971, opened 2022-09-26
Solution

Anastasiia-Khmelovska:
[DEMO LINK](https://Anastasiia-Khmelovska.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,45 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  return (
+    <div className="App">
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,45 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => getAll()
```
Githi54, line 18: separeted)

src/App.tsx
```diff
@@ -1,27 +1,45 @@
...
+        Load all goods
+      </button>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={() => get5First()
```
Githi54, line 27: and here)

src/GoodsList.tsx
```diff
@@ -8,7 +8,13 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
```
Githi54, line 10:
```suggestion
    {goods.map(({id, color, name}) => (
```

Githi54: :)

Githi54 (approved):
Well done!🦾

src/App.tsx
AlinaMatuschak: agree with @Githi54 , create separate method. But also you can compare your handlers: create one handler that can expect a callback(getAll, get5First, or getRedGoods) that will be called, and the result add to the state

AlinaMatuschak (changes requested):
fix @Githi54 comments

AlinaMatuschak (approved):

PR #973, opened 2022-09-26
Develop

NataliyaAn:
[DEMO LINK](https://NataliyaAn.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -33,7 +33,7 @@ export const App: React.FC = () => {
       <button
         type="button"
         data-cy="all-button"
-        onClick={() => handleAllGoods()}
+        onClick={handleAllGoods}
       >
         Load all goods
       </button>
@@ -41,7 +41,7 @@ export const App: React.FC = () => {
       <button
         type="button"
         data-cy="first-five-button"
-        onClick={() => handle5Goods()}
+        onClick={handle5Goods}
       >
         Load 5 first goods
       </button>
@@ -49,7 +49,7 @@ export const App: React.FC = () => {
       <button
         type="button"
         data-cy="red-button"
-        onClick={() => handleRedGoods()}
+        onClick={handleRedGoods}
       >
         Load red goods
       </button>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,50 +4,27 @@ import { GoodsList } from './GoodsList';
 import { Good } from './types/Good';
 
 import { getAll, get5First, getRedGoods } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
 
 export const App: React.FC = () => {
   const [users, setUsers] = useState<Good[]>([]);
-  // const [loudAll, setLoudAll] = useState(false);
-  // const [loudFiveGoods, setFiveGoods] = useState(false);
-  // const [loudRedGoods, setRedGoods] = useState(false);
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,60 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [users, setUsers] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleAllGoods = async () => {
+    const users1 = await getAll();
```
anna-kolmakova, line 12: I think it will be great to name const allGoods, redGoods, 5firstGoods instead of users1, users2, users 3 - for better understanding.
Also in state - const [goods, setGoods] = useState<Good[]>([]) - because we are working with goods, not users

anna-kolmakova (approved):
Good job! 👍

src/api/goods.ts
```diff
@@ -10,10 +10,17 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
+      return [...goods]
+        .sort((user1, user2) => user1.name.localeCompare(user2.name))
```
AlinaMatuschak, line 15: user1 and user2?

src/api/goods.ts
```diff
@@ -10,10 +10,17 @@ export function getAll(): Promise<Good[]> {
...
+    });
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => {
+      return [...goods]
+        .filter(user => user.color === 'red');
+    }); // get only red
```
AlinaMatuschak, line 25: delete comment

src/App.tsx
```diff
@@ -1,27 +1,60 @@
...
+  const handle5Goods = async () => {
+    const users2 = await get5First();
 
-    <GoodsList goods={[]} />
-  </div>
-);
+    setUsers(users2);
+  };
+
+  const handleRedGoods = async () => {
```
AlinaMatuschak, line 23: you can compare your handlers: create one handler that can expect a callback(getAll, get5First, or getRedGoods) that will be called, and the result add to the state

AlinaMatuschak (changes requested):
fix @anna-kolmakova comments

src/App.tsx
```diff
@@ -1,27 +1,46 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={async () => handleClick(getAll)}
```
AlinaMatuschak, line 22: you don’t use await, so don’t need async
```suggestion
        onClick={() => handleClick(getAll)}
```

src/api/goods.ts
AlinaMatuschak: comment

AlinaMatuschak (changes requested):
Good. Just small issues 👇

AlinaMatuschak (approved):
🚀

PR #976, opened 2022-09-27
react dynamic list of goods

artem-kodesnikov:
[DEMO LINK](https://artem-kodesnikov.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/GoodsList.tsx
+++ b/src/GoodsList.tsx
@@ -8,7 +8,7 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li key={good.id} data-cy="good" style={{ color: `${good.color}` }}>
         {good.name}
       </li>
     ))}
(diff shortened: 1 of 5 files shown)


src/GoodsList.tsx
```diff
@@ -8,7 +8,7 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li key={good.id} data-cy="good" style={{ color: `${good.color}` }}>
```
yungdena, line 11: You should destructure 'good' and also it's better to write each attribute on a blank line

yungdena (changes requested):
good at all, just add changes according to my comment👌

yungdena (approved):
👌

src/GoodsList.tsx
```diff
@@ -7,10 +7,14 @@ type Props = {
...
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
-        {good.name}
-      </li>
-    ))}
+    {goods.map(good => {
+      const { id, color, name } = good;
+
+      return (
+        <li key={id} data-cy="good" style={{ color: `${color}` }}>
```
AndriiRut, line 14: what typeof color?
what if pass like that?
```suggestion
        <li key={id} data-cy="good" style={{ color }}>
```

AndriiRut (approved):

PR #991, opened 2022-09-27
Develop

Maxim-Gumeniuk:
[DEMO LINK](https://Maxim-Gumeniuk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((g1, g2) => g1.name.localeCompare(g2.name))
+      .slice(0, 5));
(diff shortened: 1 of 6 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((g1, g2) => g1.name.localeCompare(g2.name))
```
erikkaraman, line 14: it's not a big mistake, but it's better to rename let g1 and g2 to good1 and good2

erikkaraman (approved):
good work👍

src/api/goods.ts
AlinaMatuschak: Agree with @erikkaraman )

AlinaMatuschak (approved):

PR #992, opened 2022-09-27
React solution: Dynamic list of goods

wfurs0w:
[DEMO LINK](https://wfurs0w.github.io/react_dynamic-list-of-goods/)


Reptiloid044 (approved):
Nice solution!

PR #1009, opened 2022-09-28
Version 0.1

David-Shahmatov:
[DEMO LINK](https://David-Shahmatov.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,52 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import 'bulma/css/bulma.min.css';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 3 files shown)


yungdena (approved):
🔥

src/api/goods.ts
```diff
@@ -10,10 +10,14 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods)
+    .then(goods => goods
+      .sort((word1, word2) => word1.name.localeCompare(word2.name))
```
dellachaise, line 15: In this case it is `good` not `word` - better to rename.

src/App.tsx
```diff
@@ -1,27 +1,70 @@
...
+        data-cy="red-button"
+        onClick={() => handleClick(getRedGoods)}
+        className="button is-danger"
+      >
+        Load red goods
+      </button>
+
+    {isVisible
+      &&
+        <div
```
dellachaise, line 57: Better use button instead of `div`. Or add changes according to errors 👇 
Screenshot 2022-09-29 at 00 20 48

dellachaise (changes requested):
Great job 👍 

Looks pretty 🔥 
And clear button is awsome 👍 

Just use `button` instead of `div`.

dellachaise (approved):
Well done 👍

PR #1012, opened 2022-09-29
Solution

Black3942:


src/App.tsx
```diff
@@ -1,27 +1,57 @@
...
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={() => {
+          getRed()
```
alexandra-protyanova, line 45: better to move this logic to separate function,  and handle possible errors

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1015, opened 2022-09-30
Dynamic list of goods

Alej4ndro1:
• [DEMO LINK](https://alej4ndro1.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,65 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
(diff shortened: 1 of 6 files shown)


kostya-kosyuk (approved):
GJ

src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const showGoods = () => {
```
AlinaMatuschak, line 13: you can compare your handlers: create one handler that can expect a callback(getAll, get5First, or getRedGoods) that will be called, and the result add to the state

src/types/Good.ts
```diff
@@ -1,4 +1,5 @@
 export interface Good {
+  tolocalecompare(b: Good): number;
```
AlinaMatuschak, line 2: why did you add it?

AlinaMatuschak (changes requested):

src/types/Good.ts
Alej4ndro1: It's not me)

AlinaMatuschak (approved):

PR #1027, opened 2022-10-12
solution

ihorandrianov:
[DEMO LINK](https://ihorandrianov.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,74 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import { useQueries } from '@tanstack/react-query';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+type SelectedGoods = 'allGoods' | 'firstFiveGoods' | 'allRedGoods';
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [goodsSelect, setGoodsSelect] = useState<SelectedGoods>('allGoods');
+  const results = useQueries({
+    queries: [
+      { queryKey: ['allGoods'], queryFn: getAll, staleTime: 0 },
+      { queryKey: ['first5'], queryFn: get5First, staleTime: 0 },
+      { queryKey: ['allRed'], queryFn: getRedGoods, staleTime: 0 },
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,74 @@
...
-    </button>
+export const App: React.FC = () => {
+  const [goodsSelect, setGoodsSelect] = useState<SelectedGoods>('allGoods');
+  const results = useQueries({
+    queries: [
+      { queryKey: ['allGoods'], queryFn: getAll, staleTime: 0 },
+      { queryKey: ['first5'], queryFn: get5First, staleTime: 0 },
+      { queryKey: ['allRed'], queryFn: getRedGoods, staleTime: 0 },
+    ],
+  });
```
Denchepornyuk, line 18: You should not make three requests at a time, it is better to make one request with one click on a button

https://user-images.githubusercontent.com/37452763/197801638-d26c58c2-e832-433d-a607-48d58e6170c6.mov

src/api/goods.ts
```diff
@@ -10,10 +10,12 @@ export function getAll(): Promise<Good[]> {
...
+    .then(goods => goods
+      .sort((first, second) => first.name.localeCompare(second.name))
+      .splice(0, 5)); // sort and get the first 5
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods.filter(good => good.color === 'red')); // get only red
 };
```
Denchepornyuk, line 21: would be great to rewrite this code with async & await syntax

src/api/goods.ts
```diff
@@ -3,17 +3,25 @@ import { Good } from '../types/Good';
...
-  return getAll()
-    .then(goods => goods); // sort and get the first 5
-};
 
-export const getRedGoods = () => {
-  return getAll()
-    .then(goods => goods); // get only red
+export const getGoods = async (query: QueryTypes) => {
+  const result = await fetch(API_URL);
+  const goodsList: Good[] = await result.json();
```
Denchepornyuk, line 15: nope, it should be three separate requests
When user clicks on get5First you should make one request and get only first five good, when user clicks getRedGoods you should make one request and get only red ones, ...
Don't make one request where you are fetching all info

Denchepornyuk (changes requested):

src/api/goods.ts
ihorandrianov: https://user-images.githubusercontent.com/16674930/198024148-8edecf2d-160f-4c57-8528-ccabd0368464.mov

Denchepornyuk: Yes, you make one request, but it is the same request for each button. Each button must have a separate function that makes a request to the server

Denchepornyuk (approved):

PR #1030, opened 2022-10-15
Develop

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,66 @@
-import React from 'react';
-import './App.scss';
+import React, { useState } from 'react';
+import { Good } from './types/Good';
 import { GoodsList } from './GoodsList';
+import './App.scss';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const goodsInitial: Good[] = [];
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
+          getAll()
+            .then((data) => {
+              setGoods(data);
+            }).catch(() => {
+              throw new Error('Error, sorry :(');
+            });
```
Denchepornyuk, line 26: let's move it to the separate method
Also, would be great to use async & await instead of then

src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={() => {
+          get5First()
+            .then((data) => {
+              setGoods(data);
+            }).catch(() => {
+              throw new Error('Error, sorry :(');
+            });
```
Denchepornyuk, line 41: the same comments here and for the `Load red goods`

src/api/goods.ts
```diff
@@ -10,10 +10,16 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
```
Denchepornyuk, line 13: yes, it was written with `then`, but you are an engineer, let's rewrite it to async & await 🙂

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1034, opened 2022-10-23
solution

Naveed-Shoukat:
[DEMO LINK](https://naveed-shoukat.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,65 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [visiableGoods, setVisiableGoods] = useState<Good[]>([]);
+  const [showGoods, setShowGoods] = useState('');
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  useEffect(() => {
+    if (showGoods === 'getAll') {
+      getAll()
+        .then(result => {
+          setVisiableGoods(result);
+        });
+    }
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+    if (showGoods === 'get5First') {
+      get5First()
+        .then(result => {
+          setVisiableGoods(result);
+        });
+    }
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+    if (showGoods === 'getRed') {
+      getRedGoods()
+        .then(result => {
+          setVisiableGoods(result);
+        });
+    }
+  }, [showGoods]);
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  return (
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [visiableGoods, setVisiableGoods] = useState<Good[]>([]);
+  const [showGoods, setShowGoods] = useState('');
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  useEffect(() => {
+    if (showGoods === 'getAll') {
```
fatihsamur, line 12: maybe switch-case?

fatihsamur (approved):
👍

src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [visiableGoods, setVisiableGoods] = useState<Good[]>([]);
+  const [showGoods, setShowGoods] = useState('');
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  useEffect(() => {
```
VinnieJ-2k20, line 11: Let's not use useEffect for this operation. You can trigger state change directly from button click.

VinnieJ-2k20 (changes requested):
Let's restructure your App a bit. use handler function (or several) instead of useEffect.

src/App.tsx
```diff
@@ -1,35 +1,33 @@
...
+      .catch(() => setShowGoods(false))
+      .finally(() => setShowGoods(true));
+  };
+
+  const handelLoadRedGoods = () => {
+    getRedGoods()
+      .then(result => setVisiableGoods(result))
+      .catch(() => setShowGoods(false))
+      .finally(() => setShowGoods(true));
+  };
```
VinnieJ-2k20, line 30: Apply useCallback to these.

VinnieJ-2k20 (changes requested):
You forgot about useCallback

src/App.tsx
```diff
@@ -1,27 +1,72 @@
...
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handelLoadRedGoods = useCallback(
+    () => {
+      getRedGoods()
+        .then(result => setVisiableGoods(result))
+        .catch(() => setShowGoods(false))
+        .finally(() => setShowGoods(true));
```
ihor-jpeg, line 36: same

src/App.tsx
```diff
@@ -1,27 +1,72 @@
...
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handelLoad5FirstGoods = useCallback(
+    () => {
+      get5First()
+        .then(result => setVisiableGoods(result))
+        .catch(() => setShowGoods(false))
+        .finally(() => setShowGoods(true));
```
ihor-jpeg, line 26: what does finally do here?
i think it will work incorrectly if there is an error.
showGoods will always be true, even if there is an error

ihor-jpeg, line 26: what does finally do here?
i think it will work incorrectly if there is an error.
showGoods will always be true, even if there is an error

ihor-jpeg (changes requested):

ihor-jpeg (approved):

PR #1041, opened 2022-10-29
Develop

nikitaovch99:
https://github.com/nikitaovch99/react_dynamic-list-of-goods

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,49 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
+    const goods = await getRedGoods();
+
+    setgoodsFromServer(goods);
+  };
+
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
+
+      <button type="button" data-cy="all-button" onClick={loadGoods}>
```
Yevheniya-Deryka, line 34: will be better to read each atr on new str

Yevheniya-Deryka (approved):
GJ!

src/App.tsx
Anna-Tkachenko: agree, fix @Yevheniya-Deryka comment

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,61 @@
...
+export const App: React.FC = () => {
+  const [goodsFromServer, setgoodsFromServer] = useState<Good[]>([]);
+  const loadGoods = async () => {
+    const goods = await getAll();
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+    setgoodsFromServer(goods);
+  };
```
Anna-Tkachenko, line 16: you can also create one general function that will accept needed handler and will call it

Anna-Tkachenko (approved):

PR #1044, opened 2022-10-30
Solution

Proph7000:
[DEMO LINK](https://proph7000.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,63 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [users, setUsers] = useState<Good[]>([]);
```
TarasHoliuk, line 9: why users? 
```suggestion
  const [goods, setGoods] = useState<Good[]>([]);
```

src/api/goods.ts
```diff
@@ -10,10 +10,10 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => [...goods].sort().slice(0, 5));
```
TarasHoliuk, line 13: `good` is an object, so you can't just do `.sort()`
add callback in the `sort` method

TarasHoliuk (changes requested):

src/App.tsx
Proph7000: I don't now...))

vlad-demchuk (approved):

PR #1049, opened 2022-10-31
commit1

Gintako666:
[DEMO LINK](https://Gintako666.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,49 @@
-import React from 'react';
+import { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App = () => {
+  const [listGoods, setListGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
+          getAll().then(a => setListGoods(a));
+        }}
+      >
+        Load all goods
+      </button>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={() => {
+          get5First().then(a => setListGoods(a));
+        }}
+      >
+        Load 5 first goods
+      </button>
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
+          getAll().then(a => setListGoods(a));
```
alexandra-protyanova, line 21: use more meaningful names for variables

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={() => {
+          getRedGoods().then(a => setListGoods(a));
```
alexandra-protyanova, line 41: let us move it from onClick to a separate function and will think about case of error

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
+        }}
+      >
+        Load red goods
+      </button>
+      {listGoods.length > 0 && <GoodsList goods={listGoods} />}
```
alexandra-protyanova, line 46:
```suggestion
      {listGoods.length && <GoodsList goods={listGoods} />}
```

alexandra-protyanova: it is converting to boolean anyway if it will be 0, it will be false

alexandra-protyanova (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,57 @@
...
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
+          getAll().then(Goods => setListGoods(Goods));
```
DarkMistyRoom, line 29:
```suggestion
          getAll().then(goods => setListGoods(goods));
```

src/App.tsx
DarkMistyRoom: Just remove it

src/api/goods.ts
```diff
@@ -10,10 +10,12 @@ export function getAll(): Promise<Good[]> {
```
DarkMistyRoom, line 10: Let us think about a situation when an error comes from the server. The best way is to throw the error

src/App.tsx
```diff
@@ -1,27 +1,57 @@
...
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  function redGoods(goods: Good[]) {
+    if (goods.length > 0) {
+      setListGoods(goods);
+    } else {
+      setListGoods([]);
+    }
+  }
```
DarkMistyRoom, line 19: Make one handler for all buttons

DarkMistyRoom (changes requested):

src/api/goods.ts
alexandra-protyanova: this comment, wasn't fixed

alexandra-protyanova (approved):

PR #1053, opened 2022-10-31
Implemented the dynamic list of goods

anastasiia-nahaiska:
[DEMO LINK](https://anastasiia-nahaiska.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,66 @@
-import React from 'react';
+import React, { useState } from 'react';
+import 'bulma/css/bulma.css';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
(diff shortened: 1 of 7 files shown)


AntonVaida (approved):

src/App.tsx
```diff
@@ -1,27 +1,66 @@
...
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleAllGoods = () => {
+    getAll()
+      .then(goodsFromServer => {
+        setGoods(goodsFromServer);
+      });
```
Anna-Tkachenko, line 15: you repeat this logic 3 times, try to optimize it
you can create one function that will accept needed handler, make a needed request and set data to state

Anna-Tkachenko (approved):

src/App.tsx
anastasiia-nahaiska: Thanks) already done))

PR #1073, opened 2022-11-01
implemented the request to API

SerhiiRomaniukWD:
[DEMO LINK](https://serhiiromaniukwd.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,9 +13,9 @@ export const App: React.FC = () => {
   const [goods, setGoods] = useState<Good[]>([]);
 
   const loadGoodsFromServer = async (
-    promiseFromServer: () => Promise<Good[]>,
+    goodsFromServer: () => Promise<Good[]>,
   ) => {
-    setGoods(await promiseFromServer());
+    setGoods(await goodsFromServer());
   };
 
   return (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,14 +12,12 @@ import {
 export const App: React.FC = () => {
   const [goods, setGoods] = useState<Good[]>([]);
 
-  const loadGoodsFromServer = (promiseFromServer: () => Promise<Good[]>) => {
-    const setGoodsFromServer = async () => {
-      const goodsFromServer = await promiseFromServer();
+  const loadGoodsFromServer = async (
+    promiseFromServer: () => Promise<Good[]>,
+  ) => {
+    const goodsFromServer = await promiseFromServer();
 
-      setGoods(goodsFromServer);
-    };
-
-    return setGoodsFromServer();
+    setGoods(goodsFromServer);
   };
 
   return (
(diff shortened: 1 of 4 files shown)


SerhiiRomaniukWD:
I was left alone, I wrote to the group several times, but without results, that's why I'm without review

src/App.tsx
```diff
@@ -1,27 +1,52 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadGoodsFromServer = async (
+    promiseFromServer: () => Promise<Good[]>,
```
Anna-Tkachenko, line 16: will be better to create more understandable variable name: it's not promise from server, it's function that should create corresponding request to the server

src/api/goods.ts
```diff
@@ -3,17 +3,48 @@ import { Good } from '../types/Good';
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
+const request = async (url: string) => {
```
Anna-Tkachenko, line 6: will be better to add types for returned value

src/api/goods.ts
```diff
@@ -3,17 +3,48 @@ import { Good } from '../types/Good';
...
+  try {
+    const response = await fetch(url);
+
+    return await response.json();
```
Anna-Tkachenko, line 10: await is redundant here

src/api/goods.ts
```diff
@@ -3,17 +3,48 @@ import { Good } from '../types/Good';
...
+  };
+
+  return answerFromServer();
 }
 
 export const get5First = () => {
-  return getAll()
-    .then(goods => goods); // sort and get the first 5
+  const answerFromServer = async () => {
+    const jsonData = await request(API_URL);
```
Anna-Tkachenko, line 32: why json data? you converted json on js object via json() function

Anna-Tkachenko (changes requested):

src/api/goods.ts
SerhiiRomaniukWD: i will fix it

src/api/goods.ts
SerhiiRomaniukWD: it's not working without await

src/api/goods.ts
SerhiiRomaniukWD: I will fix it

src/App.tsx
SerhiiRomaniukWD: you are right

src/api/goods.ts
```diff
@@ -3,17 +3,54 @@ import { Good } from '../types/Good';
...
+    throw error;
+  }
 };
 
-export const getRedGoods = () => {
-  return getAll()
-    .then(goods => goods); // get only red
+export const get5First = async () => {
+  try {
+    const goodsFromServer = await request(API_URL);
```
Anna-Tkachenko, line 32: you can call here getAll() function

Anna-Tkachenko (approved):

PR #1075, opened 2022-11-01
implement dynamic list of goods (react)

anton-iskryk:
[DEMO LINK](https://anton-iskryk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,59 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,59 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const getAllGoods = async () => {
+    const items = await getAll();
```
artem-nasadchuk, line 11: as for me maybe better naming will be 'data' or 'goodsFromServer'

artem-nasadchuk (approved):

src/App.tsx
```diff
@@ -1,27 +1,59 @@
...
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+    setGoods(items);
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const getFiveGoods = async () => {
```
Anna-Tkachenko, line 16: you can also create one general function, that will accept needed handler and call it

Anna-Tkachenko (approved):

PR #1078, opened 2022-11-01
React dynamic list of goods

glugovsky:
[DEMO LINK](https://glugovsky.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,50 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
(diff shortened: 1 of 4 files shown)


VovaDzhuryk (approved):
GJ!

src/api/goods.ts
```diff
@@ -5,15 +5,23 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
+    .then(response => {
+      return response.json();
+    });
```
OleksiiBlezniuk, line 10: no need to use `return` keyword if you can return value after string

src/App.tsx
```diff
@@ -1,27 +1,50 @@
...
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const setGoodsFromServer = (getGoodsFromServer: () => Promise<Good[]>) => {
+    getGoodsFromServer()
+      .then((goodsFromServer: Good[]) => setGoods(goodsFromServer))
```
OleksiiBlezniuk, line 13: I think type defining can be skipped here, 'cause we already know what this function will return

src/App.tsx
```diff
@@ -1,27 +1,50 @@
...
+      .catch(error => {
+        throw new Error(error);
+      });
```
OleksiiBlezniuk, line 16: Why do you need to throw an error inside catch block? I mean, it's good to check if needed values were returned from api (to handle cases when data is returned from api with error), but just to throw same error has no sense

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
👍

PR #1079, opened 2022-11-01
dynamic-list-of-goods

vallerii:
[DEMO LINK](https://vallerii.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,63 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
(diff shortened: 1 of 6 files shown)


Oleksandra-Shevchenko (approved):
GJ 👍

src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleGetAllSubmit = () => {
+    getAll().then(goodsFromServ => {
```
Anna-Tkachenko, line 14: don't make variable names shortened, it can affect in bad way on code readability
use full variable names

src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const handleGetRedSubmit = () => {
+    getRedGoods().then(goodsFromServ => {
+      setGoods(goodsFromServ);
+    });
+  };
```
Anna-Tkachenko, line 29: will be better to create one general function that will accept needed method for making server request and then store the result in state for avoiding code repeating

Anna-Tkachenko (approved):

PR #1083, opened 2022-11-02
solution

MAKSOti:
[DEMO LINK](https://MAKSOTi.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,62 @@
-import React from 'react';
+/* eslint-disable no-console */
+import React, { useState } from 'react';
 import './App.scss';
+import 'bulma';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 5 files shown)


artem-musii (approved):
GJ!

src/App.tsx
```diff
@@ -1,27 +1,62 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goodsFromServer, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const getAllGoods = () => {
```
Anna-Tkachenko, line 13: you can also create one main function that will accept needed callback as an argument to avoid repeating yourself

src/api/goods.ts
```diff
@@ -10,10 +10,15 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods.sort((g1, g2) => (
+      g1.name.localeCompare(g2.name)
+    )))
+    .then(goods => goods.slice(0, 5));
```
Anna-Tkachenko, line 16: last then is redundant here, you can call slice method in chain after sort one

Anna-Tkachenko (approved):

PR #1089, opened 2022-11-04
react dynamic list of goods implementation

varkovroma:
[DEMO LINK](https://varkovroma.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,64 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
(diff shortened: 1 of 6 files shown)


Oleksandra-Shevchenko (approved):
👍

src/App.tsx
```diff
@@ -1,27 +1,64 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleLoadGoods = (loadButton: LoadButton) => {
```
Anna-Tkachenko, line 12: will be better just accept needed function, that should be called and then store the result of the function calling in state for avoiding code repeating

Anna-Tkachenko (approved):

PR #1096, opened 2022-11-08
solution

Andrii-Kuzmenko:
[DEMO LINK](https://Andrii-Kuzmenko.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,25 +5,18 @@ import { getAll, get5First, getRedGoods } from './api/goods';
 import { Good } from './types/Good';
 
 export const App: React.FC = () => {
-  const defaultGoods: Good[] = [];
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,27 +1,60 @@
...
+    });
+  };
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const get5Here = () => {
+    get5First().then(good => {
+      setGoods(good);
+    });
```
TarasHoliuk, line 20:
```suggestion
    get5First().then(setGoods);
```
Since `setGoods` accepts good as argument (exactly what your promise returns) you can just pass it as callback into `then`

TarasHoliuk (changes requested):

tetiana-ivanchenko (approved):

PR #1104, opened 2022-11-19
Solution

Fispil:
[DEMO LINK](https://Fispil.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((firstGood, secondGood) => firstGood.name
+        .localeCompare(secondGood.name))
+      .filter((good, index) => good && index < 5));// sort and get the first 5
 };
 
 export const getRedGoods = () => {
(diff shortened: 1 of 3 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((firstGood, secondGood) => firstGood.name
```
polosanya, line 14:
```suggestion
    .then(goods => [...goods]
      .sort((firstGood, secondGood) => firstGood.name
```
Sort is a mutable method, so it's better to make a copy

src/api/goods.ts
```diff
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
...
+        .localeCompare(secondGood.name))
+      .filter((good, index) => good && index < 5));// sort and get the first 5
```
polosanya, line 16: No need for filter, you can just slice 5 first items of array

polosanya (changes requested):
Almost done! Please check the comments to improve your code

TarasHoliuk (approved):

PR #1111, opened 2022-11-22
solution

GizemEce:
[DEMO LINK](https://GizemEce.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,50 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [allGoods, setAllGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadGoods = async (sortType: Promise<Good[]>) => {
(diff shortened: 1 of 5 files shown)


src/GoodsList.tsx
```diff
@@ -5,10 +5,20 @@ type Props = {
...
+};
+
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li
+        key={good.id}
+        data-cy="good"
+        style={style(good.color)}
```
hajnaloltyan, line 20: it can be added in a simpler way like style={{ color: good.color }}

hajnaloltyan (approved):
Great work!

src/App.tsx
```diff
@@ -1,27 +1,50 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [allGoods, setAllGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadGoods = async (sortType: Promise<Good[]>) => {
```
VinnieJ-2k20, line 13: Apply useCallback here.

VinnieJ-2k20, line 13: Don't use Promise and function argument. Either pass just the callback here and call it inside the function, or resolve it before and pass only result here.

src/GoodsList.tsx
VinnieJ-2k20: ^^^ Good suggestion, operation is simple enough to avoid creating extra callbacks.

VinnieJ-2k20 (changes requested):
Your solution works, but some improvements needed.

VinnieJ-2k20 (approved):

Andr1yK (changes requested):
test

Andr1yK (approved):

PR #1118, opened 2022-11-26
solution

1stIndominus:
[DEMO LINK](https://1stIndominus.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,45 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,45 @@
...
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const loadHandler = async (loader: () => Promise<Good[]>) => (
+    setGoods(await loader())
+  );
```
Anastasiia-Svintsova, line 12: It's better to move the result of the async function in a variable and use it in the setState function
```suggestion
  const loadHandler = async (loader: () => Promise<Good[]>) => (
    const goodsFromServer = await loader();
    setGoods(goodsFromServer)
  );
```

Anastasiia-Svintsova (approved):

PR #1127, opened 2022-12-05
Render dynamic list with data from server

anastasiia-tilikina:
[DEMO LINK](https://anastasiia-tilikina.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,49 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+type GoodsStateValue = [] | Good[];
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState([] as GoodsStateValue);
 
-    <button type="button" data-cy="first-five-button">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState([] as GoodsStateValue);
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const showGoods = (getFn: () => Promise<Good[]>) => {
```
vlad-demchuk, line 13: let's move `getFn` type to a separate `type`

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+type GoodsStateValue = [] | Good[];
```
vlad-demchuk, line 8: when you use array you can set just `Good[]`. even if your array is empty it will work

vlad-demchuk: remove this type and set type for state

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState([] as GoodsStateValue);
```
vlad-demchuk, line 11:
```suggestion
  const [visibleGoods, setVisibleGoods] = useState<Good[]>([]);
```

vlad-demchuk (changes requested):

TarasHoliuk (approved):

PR #1128, opened 2022-12-06
Develop

andriy-shymkiv:
[DEMO LINK](https://andriy-shymkiv.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,65 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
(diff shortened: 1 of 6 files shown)


Anna-Shapovalova (approved):
Cool!) I have the same)

src/App.tsx
```diff
@@ -1,27 +1,50 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleClick = async (callback: Promise<Good[]>) => {
+    setGoods(await callback);
```
illiarerun, line 12: I don't like this await in function call, better create variable for that

illiarerun: And also, it's not really a callback, it's a result of your function, so better rename it

PR #1130, opened 2022-12-06
Dynamic list of goods was added

iryna-lelyk:
[DEMO LINK](https://iryna-lelyk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,48 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRed } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [loadedGoods, setLoadedGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
+        type="button"
+        data-cy="red-button"
+        onClick={() => getRed().then(redGoods => (
+          setLoadedGoods(redGoods)
+        ))}
+      >
+        Load red goods
+      </button>
+
+      <GoodsList goods={loadedGoods} />
```
denys-danyliuk, line 45: if you don't have any goods you don't need to render this component

denys-danyliuk (changes requested):
🔥

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
+
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={() => loadGoods(getRed())}
+      >
+        Load red goods
+      </button>
+
+      {loadedGoods.length > 0 && (
```
denys-danyliuk, line 44:
```suggestion
      {loadedGoods.length && (
```

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [loadedGoods, setLoadedGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadGoods = (getGoods: Promise<Good[]>) => {
+    getGoods.then(goods => (
+      setLoadedGoods(goods)));
```
denys-danyliuk, line 13:
```suggestion
    getGoods.then(setLoadedGoods));
```
it will work this way

denys-danyliuk (approved):

src/App.tsx
iryna-lelyk: If I do write it like that, it will display 0 on the page, when no goods are showed.

PR #1141, opened 2022-12-07
solution

Oleksandr-Stryzhavchuk:
[DEMO LINK](https://Oleksandr-Stryzhavchuk.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,27 +1,57 @@
...
+      .then(resultFive => setGoods(resultFive));
+  }
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  function getRed() {
+    getAllRed()
+      .then(resultRed => setGoods(resultRed));
+  }
```
andrii-bassarab, line 24: you are duplicating the logic several times, it is better to put it in a separate function and pass different arrays to its parameter when you call it for different buttons

mykola-koval (approved):

PR #1161, opened 2022-12-13
Done

ruslik87:
[DEMO LINK](https://ruslik87.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,44 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,44 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => getAll().then(setSortGoods)}
```
Bohdan-Maksymiuk, line 18: lets make it in more common way
create handler function, which obtain callback (one of these `getAll, get5First, getRedGoods`), call it and set state

Bohdan-Maksymiuk (changes requested):
Screenshot 2022-12-14 at 15 47 04

SanyaBratashchuk (approved):

PR #1163, opened 2022-12-16
Develop

p-omogit-e:
[DEMO LINK](https://p-omogit-e.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,66 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { Good } from './types/Good';
+import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const getAllGoods = async () => {
+    const loadedGoods = await goodsAPI.getAll();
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,66 @@
...
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
+
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
+          getAllGoods();
+        }}
```
vlad-demchuk, line 38: you can not to create a callback if you no need to pass arguments
```suggestion
        onClick={getAllGoods}
```

src/api/goods.ts
```diff
@@ -8,12 +8,28 @@ export function getAll(): Promise<Good[]> {
     .then(response => response.json());
 }
 
-export const get5First = () => {
-  return getAll()
-    .then(goods => goods); // sort and get the first 5
+export const get5First = async () => {
+  const loadedGoods = await getAll()
+    .then(goods => goods);
```
vlad-demchuk, line 13: don't mix `async/await/try/catch` with `then/catch` in one project

src/api/goods.ts
```diff
@@ -8,12 +8,28 @@ export function getAll(): Promise<Good[]> {
...
+    if (a.name > b.name) {
+      return 1;
+    }
+
+    if (a.name < b.name) {
+      return -1;
+    }
+
+    return 0;
+  });
```
vlad-demchuk, line 25: it's not working correctly for strings. use `localeCompare`

vlad-demchuk (changes requested):

src/api/goods.ts
p-omogit-e: How can I do sort without use async? "Property 'sort' does not exist on type 'Promise<Good[]>'.ts(2339)
goods.ts(15, 15): Did you forget to use 'await'?"

nazarmatsevych (approved):

nazarmatsevych (approved):

PR #1166, opened 2022-12-20
Dynamic list of goods task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,84 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [clickedButton, setClickedButton] = useState('');
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadGoods = async () => {
+    try {
+      switch (clickedButton) {
+        case 'all':
+          setGoods(await getAll());
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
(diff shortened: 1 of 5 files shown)


juliiaap:
All tests passed, when running locally in VS Code. But, PR run failed with error in the last test.
dynamic list of goods tests

src/App.tsx
```diff
@@ -1,27 +1,84 @@
...
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
+
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
+          setClickedButton('all');
+          loadGoods();
```
polosanya, line 53: I think it's better to create 3 separate functions in order of trying to handle everything in one. 
It would make your code cleaner and more understandable. It's an interesting solution, but loadGoods function looks too big and complicated

src/GoodsList.tsx
```diff
@@ -8,7 +8,7 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li key={good.id} data-cy="good" style={{ color: good.color }}>
```
polosanya, line 11:
```suggestion
      <li 
        key={good.id} 
        data-cy="good" 
        style={{ color: good.color }}
      >
```

polosanya (changes requested):
Almost perfect!

src/GoodsList.tsx
juliiaap: fixed

src/App.tsx
juliiaap: done

src/App.tsx
```diff
@@ -1,27 +1,136 @@
...
+  // const onButtonClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+  //   if (clickedButton !== event.currentTarget.id) {
+  //     setClickedButton(event.currentTarget.id);
+  //     loadGoods();
+  //   }
+  // };
+
+  // useEffect(() => {
+  //   loadGoods();
+  // }, [clickedButton]);
```
qoa11a, line 79: Remove unnecessary comments.

qoa11a (approved):

src/App.tsx
juliiaap: removed

PR #1177, opened 2023-01-02
react_dynamic-list-of-goods

ysrckr:
[DEMO LINK](https://ysrckr.github.io/react_dynamic-list-of-goods/)


UmizDemud (approved):
👍  I should dig deeper, can't just approve all your code 😄 just seems correct from all angles

PR #1185, opened 2023-01-07
Solution

Julia2063:
[DEMO LINK](https://Julia2063.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,53 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+export const App: React.FC = () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
-  </div>
-);
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
+
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => loadGoods(getAll())}
```
Anastasiia-Svintsova, line 29: Don't call the function here. Call it inside loadGoods
```suggestion
        onClick={() => loadGoods(getAll)}
```

src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
-    </button>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const loadGoods = async (callback: Promise<Good[]>) => {
+    try {
+      const loadedGoods = await callback;
```
Anastasiia-Svintsova, line 14:
```suggestion
      const loadedGoods = await callback();
```

src/types/Good.ts
```diff
@@ -1,4 +1,5 @@
 export interface Good {
+  localeCompare(g2: Good): number;
```
Anastasiia-Svintsova, line 2: Do you really need it in the interface? I think you can remove it

Anastasiia-Svintsova (changes requested):
Good job! Fix some comments, please

qoa11a (approved):

PR #1188, opened 2023-01-10
solution

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,27 +1,62 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => loadGoods(getAll, Load.All)}
```
UmizDemud, line 36: The logic for the first parameter can be moved to inside of the loadGoods function

src/App.tsx
```diff
@@ -1,27 +1,62 @@
...
+        data-cy="red-button"
+        onClick={() => loadGoods(getRedGoods, Load.Red)}
+      >
+        Load red goods
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      {goods.length > 0 && (
```
UmizDemud, line 57: this can be 

```js
{goods.length && (

UmizDemud (approved):
I went through twice. Just two comments that aren't crucial. Approving, so that you can decide whether they make sense

src/App.tsx
lefteris-L: It needs to be explicit boolean cause in some cases the number shows up next to the element

naziamoff (approved):
nice job 👍

PR #1192, opened 2023-01-12
react_dynamic-list-of-goods

kolya-movchan:
[DEMO LINK](https://kolya-movchan.github.io/react_dynamic-list-of-goods/)


src/App.scss
```diff
@@ -1 +1,27 @@
...
+li {
+  margin: 0;
+}
+
+button {
+  padding: 10px;
+  width: 150px;
+  background-color: lightblue;
+  margin-bottom: 10px;
+}
```
tetiana-ivanchenko, line 22: don't use tags as selectors, add classes

tetiana-ivanchenko (changes requested):

src/App.scss
```diff
@@ -1 +1,27 @@
-// styles go here
+body {
+  display: flex;
+  justify-content: center;
+  text-align: center;
+}
```
vlad-demchuk, line 5: check the previous comment

vlad-demchuk (changes requested):

nazarmatsevych (approved):

PR #1199, opened 2023-01-16
solution

Sveta-Kryukova:
[DEMO LINK](https://Sveta-Kryukova.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,57 @@
-import React from 'react';
 import './App.scss';
+import { useState } from 'react';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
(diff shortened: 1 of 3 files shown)


DiDemkiv (approved):
Everything is fine, but you have to add some styles:) 👍

Ihor-Karpyn (changes requested):
Your code is fine but UI looks like Wikipedia, let's improve it

src/App.tsx
```diff
@@ -1,27 +1,58 @@
...
+      .then(setGoods);
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const handleLoaderOfRedGoods = () => {
+    getRedGoods()
+      .then(setGoods);
+  };
```
solaryasha, line 24: What do you think about creating one callback that accepts the function?
Seems like your functions do the same:
1. calling the function which returns the promise
2. pass `setGoods` function inside `then` method

solaryasha (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleLoader = (loadedGoods: Promise<Good[]>) => {
```
solaryasha, line 11:
```suggestion
  const handleLoadGoods = (loadedGoods: Promise<Good[]>) => {
```

solaryasha (approved):
👍

PR #1203, opened 2023-01-17
solution

DimaKolesnyk:
[DEMO LINK](https://dimakolesnyk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,69 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import 'bulma/css/bulma.css';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleLoadAllGoodsClick = () => {
+    getAll()
+      .then(AllGoods => setGoods(AllGoods));
+  };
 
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -1,27 +1,70 @@
...
+  const handleFirst5GoodsClick = async () => {
+    setGoods(await get5First());
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handleRedGoodsClick = async () => {
+    setGoods(await getRedGoods());
+  };
```
RedkoSerhii, line 22: you can use one function instead of three, like this:
  const handleButtonSubmit = async (callback: Promise<Good[]>) => {
    setVisibleGoods(await callback);

and call it: onClick={() => handleButtonSubmit(getAll())}

RedkoSerhii (changes requested):

src/App.tsx
RedkoSerhii: will be better use another name for callback))  like loadedGoods, for example

RedkoSerhii (approved):
GJ!

src/App.tsx
```diff
@@ -1,27 +1,61 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleButtonSubmit = async (callback: Promise<Good[]>) => {
```
valerii-derkach, line 11:
```suggestion
  const handleButtonSubmit = async (loadedGoods: Promise<Good[]>) => {
```

valerii-derkach (changes requested):

src/App.tsx
DimaKolesnyk: Oh, my bad, ty

valerii-derkach (approved):

PR #1204, opened 2023-01-17
solution

Fynkzia:
[DEMO LINK](https://Fynkzia.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,14 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => (
+      [...goods]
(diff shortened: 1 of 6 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,14 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => (
+      [...goods]
```
batstolya, line 14: you don't need to make copy array, because each time you click on the button, you will receive a new array, so in fact sorting will not change the original array

batstolya (approved):
good job 🔥

src/api/goods.ts
solaryasha: I agree with @batstolya. We don't need to make the copy

solaryasha (approved):
Nice !

PR #1209, opened 2023-01-18
first commit

vladimir-arkanov:
[DEMO LINK](https://vladimir-arkanov.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
+
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+
+  const handleListOfGoods = () => {
+    getAll()
+      .then((good) => {
+        setGoods(good);
+      });
+  };
```
Sveta-Kryukova, line 20: you can make it shorter 
const handleListOfGoods = () => {
  getAll()
    .then(setGoods);
};

src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
+      .then((goodName) => {
+        setGoods(goodName);
+      });
+  };
+
+  const handleSortColorRed = () => {
+    getRedGoods().then((goodColor) => {
+      setGoods(goodColor);
+    });
+  };
```
Sveta-Kryukova, line 33: here is also you can make it shorter

Sveta-Kryukova (changes requested):
Good job)

src/App.tsx
vladimir-arkanov: i've' done only for better understanding and education as Vova said

valerii-derkach (approved):

PR #1214, opened 2023-01-18
Solution

Yuriy-Khoptyanii:
[DEMO LINK](https://Yuriy Khoptyanii.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2090,9 +2090,9 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "1.2.3",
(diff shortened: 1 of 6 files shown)


RedkoSerhii (approved):
GJ, but will be better add some stiles.
And I am not shure, that "loadGoods" is good name. Like "handle.." will be better.

valerii-derkach (changes requested):
STYLES

valerii-derkach (approved):

PR #1223, opened 2023-01-18
react_dynamic-list-of-goods

Mykhailyk-Vadym:
[DEMO LINK](https://Mykhailyk-Vadym.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 import { Good } from './types/Good';
+import 'bulma';
 
 import { getAll, get5First, getRedGoods } from './api/goods';
 // or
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,52 @@
...
-    </button>
+  return (
+    <div className="App">
+      <h1 className="title is-1">Dynamic list of Goods</h1>
+      <div className="buttons is-centered">
+        <button
+          type="button"
+          className="button is-info is-rounded is-light"
+          data-cy="all-button"
+          onClick={() => (getAll().then(good => setGoods(good)))}
```
NadiiaKoch, line 25: Interesting! But it's good to set data in a separate function

NadiiaKoch (approved):
Good!

src/App.tsx
solaryasha: I agree with @NadiiaKoch. It's a valid comment.

src/App.tsx
```diff
@@ -1,27 +1,52 @@
...
+        </button>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+        <button
+          type="button"
+          className="button is-link is-rounded is-light"
+          data-cy="first-five-button"
+          onClick={() => (get5First().then(good => setGoods(good)))}
```
solaryasha, line 34: In all three `onClick` we basically do the same.
Call some function and use `then` for adding data to the state.
Let's try to create a function that takes a function, call it, and add data to the state

solaryasha (changes requested):

src/App.tsx
Mykhailyk-Vadym: ok. i did it before but then sought that we can write it in 1 row so i thought why not?

src/App.tsx
```diff
@@ -1,27 +1,56 @@
...
-    </button>
+  // 'Saw this method when did studen review'
+  // const handleButtonClick = async (callback: Promise<Good[]>) => {
+  //   setGoods(await callback);
+  // };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handleButtonClick = (func: Promise<Good[]>) => (
```
solaryasha, line 17: It's not exactly function, it's the result of called function 
```suggestion
  const handleButtonClick = (loadedGoods: Promise<Good[]>) => (
```

solaryasha (approved):
👍

PR #1232, opened 2023-01-19
solution

OlexandraKorol:
[DEMO LINK](https://OlexandraKorol.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,60 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { get5First, getAll, getRedGoods } from './api/goods';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import 'bulma/css/bulma.css';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,60 @@
...
+  const handleClickAll = () => {
+    getAll().then((data) => setGoods(data))
+  }
 
-    <GoodsList goods={[]} />
-  </div>
-);
+  const handleClickMax5 = () => {
+    get5First().then((data) => setGoods(data))
+  }
```
Sveta-Kryukova, line 21: you can make it shorter)
  const handleClickMax5 = () => {
    get5First().then(setGoods)
  }

src/App.tsx
```diff
@@ -1,27 +1,60 @@
...
+  const handleClickAllRed = () => {
+    getRedGoods().then((data) => setGoods(data))
+  }
```
Sveta-Kryukova, line 24: here is also)

Sveta-Kryukova (changes requested):
Also every of your good should be colored (in styles) with the same color as in api) check this in readme file)

Ihor-Karpyn (approved):
- list padding
image

GJ but pay more attention to UI

PR #1236, opened 2023-01-19
solution

DiDemkiv:
[DEMO LINK](https://DiDemkiv.github.io/react_dynamic-list-of-goods/)


Sveta-Kryukova (approved):

Ihor-Karpyn (changes requested):
Your code is ok but UI looks like Wikipedia, let's fix this issue

PR #1245, opened 2023-01-25
Solution

CharAnagn:
[DEMO LINK](https://charanagn.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,48 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={async () => {
+          setGoods(await getAll());
+        }}
```
ysrckr, line 20: Wouldn't it look better if it was in a function?

src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={async () => {
+          setGoods(await get5First());
+        }}
```
ysrckr, line 30: Same here

ysrckr (approved):
Just a small change make it perfect. onClicks with async function inside doesn't look well.

src/App.tsx
naziamoff: indeed, @ysrckr is right

naziamoff (changes requested):
already looks good, but let's create handlers for the buttons

naziamoff (approved):

PR #1254, opened 2023-01-30
Develop

Grayfox147:
[DEMO LINK](https://Grayfox147.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -11,6 +11,8 @@ import { Good } from './types/Good';
 export const App: React.FC = () => {
   const [goods, setGoods] = useState<Good[]>([]);
   const isEmpty = goods.length === 0;
+  const arentRed = goods.some((good) => good.color !== 'red');
+  const areRed = goods.every((good) => good.color === 'red');
 
   const handleLoadAllGoods = async () => {
     const loadedGoods = await getAll();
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+    return (isEmpty || goods.length === 5)
+      ? (
+        setGoods(loadedGoods)
+      )
+      : (
+        setGoods([])
+      );
```
meforoma, line 21: i'm confused why would you need all these
just set your state to loadedgoods and that's it
```suggestion
    setGoods(loadedGoods);
```

src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+    if (isEmpty || goods.length > 5) {
+      setGoods(loadedGoods);
+    } else if (goods.length === 5) {
+      if (!areRed) {
+        setGoods([]);
+      }
```
meforoma, line 32:
```suggestion
    setGoods(loadedGoods);
```

src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
+  const handlerRedGoods = async () => {
+    const loadedGoods = await getRedGoods();
+
+    return (isEmpty || !areRed)
+      ? (
+        setGoods(loadedGoods)
+      )
+      : (
+        setGoods([])
+      );
```
meforoma, line 47:
```suggestion
    setGoods(loadedGoods);
```

meforoma (changes requested):
i'm not sure i understand those conditionals in handler functions 🤔

src/App.tsx
Grayfox147: Ah I was just playing that they would work together the bottons and also like if I click again they would hide the list! etc, I will remove all the extra and commit again!

meforoma (approved):
✅

PR #1256, opened 2023-01-30
react_dynamic-list-of-goods

JzSol:
[DEMO LINK](https://JzSol.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .slice(0, 5)); // sort and get the first 5
 };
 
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,15 @@ export function getAll(): Promise<Good[]> {
...
+      .slice(0, 5));
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+  // eslint-disable-next-line
+    .then(goods => goods
+      // eslint-disable-next-line
+      .filter(goods => goods.color === 'red'));
```
meforoma, line 23: why do you need to disable eslint?
i'm sure you can avoid it

meforoma, line 23: i guess parenthesis would satisfy eslint
```suggestion
    .then((goods) => goods
      .filter((goods) => goods.color === 'red'));
```

meforoma (approved):
it's a score
but i insist you don't disable eslint without a good reason.

PR #1262, opened 2023-01-30
solution

Iksich:


src/api/goods.ts
```diff
@@ -10,10 +10,25 @@ export function getAll(): Promise<Good[]> {
...
+        if (nameA < nameB) {
+          return -1;
+        }
+
+        if (nameA > nameB) {
+          return 1;
+        }
+
+        return 0;
+      })
```
meforoma, line 27: just use the localeCompare
```suggestion
      .sort((a, b) => a.name.localeCompare(b.name))
```

meforoma (approved):
✅ 
but your sort is over-engineered 😉

PR #1269, opened 2023-02-06
solution

igor-kashchenko:
[DEMO LINK](https://igor-kashchenko.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,60 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleAll = async () => {
+    const allGoods = await getAll();
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+    setGoods(allGoods);
+  };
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,60 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleAll = async () => {
+    const allGoods = await getAll();
```
denys-cheporniuk, line 12: what if getAll will throw an error, for example, a server isn't available?
You should catch errors and handle it, page should work if request isn't successful

src/api/goods.ts
```diff
@@ -10,10 +10,11 @@ export function getAll(): Promise<Good[]> {
...
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5));
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods.filter(good => good.color === 'red'));
 };
```
denys-cheporniuk, line 20: use async await instead of then

denys-cheporniuk (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,85 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleAll = async () => {
```
Anna-Tkachenko, line 12: you can also create one unique function that can accept function for getting data from api as an argument

Anna-Tkachenko: it will help to avoid repeating your code

Anna-Tkachenko (approved):
Good job!

Anna-Tkachenko (changes requested):
check point 1 in checklist

Anna-Tkachenko (approved):
⭐️

PR #1270, opened 2023-02-06
react_dynamic-list-of-goods

DragotaIV:
https://DragotaIV.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,14 @@
-// styles go here
+ul {
+  list-style: none;
+  padding: 0;
+  width: 40vw;
+  margin-left: 30px;
+}
+
+li {
+  margin: 5px 0;
+}
+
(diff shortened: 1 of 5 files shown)


tetiana-ivanchenko (changes requested):
goods should be different colors

src/App.scss
```diff
@@ -1 +1,32 @@
...
+    background-color: green;
+  }
+
+  &--five {
+    background-color: orange;
+  }
+
+  &--red {
+    background-color: red;
+  }
```
tetiana-ivanchenko, line 30: why do you need this classes?

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1272, opened 2023-02-08
Dynamic list of goods with API call

Kavster92:
[DEMO LINK](https://Kavster92.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,15 +1,29 @@
+/* eslint-disable max-len */
 import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
-import * as goodsFilterFunctionAPI from './api/goods';
+import { getAll, get5First, getRedGoods, getSelectedColor } from './api/goods';
 import { Good } from './types/Good';
 
 export const App: React.FC = () => {
   const [goods, setGoods] = useState<Good[]>([]);
 
-  const buttonClickHandle = async (loadedGoods: Promise<Good[]>) => (
-    setGoods(await loadedGoods)
-  );
+  const getAllGoods = async () => {
+    setGoods(await getAll());
+  };
+
+  const getFirstFiveGoods = async () => {
+    setGoods(await get5First());
+  };
+
+  const setRedGoods = async () => {
+    setGoods(await getRedGoods());
+  };
+
+  const getSelectedColoredGoods = async (color: string) => {
+    setGoods(await getSelectedColor(color));
+  };
+
 
   return (
 
@@ -20,7 +34,7 @@ export const App: React.FC = () => {
         className="button"
         type="button"
         data-cy="all-button"
-        onClick={() => buttonClickHandle(goodsFilterFunctionAPI.getAll())}
+        onClick={getAllGoods}
       >
         Load All Goods
       </button>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleClick = async (loadedGoods: Promise<Good[]>) => (
+    setGoods(await loadedGoods));
```
naziamoff, line 11:
```suggestion
  const handleClick = async (loadedGoods: Promise<Good[]>) => (
    setGoods(await loadedGoods)
  );
```

naziamoff: but handle what kind of click? This name does not give me any additional information

src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
+        onClick={() => handleClick(goodsAPI.get5First())}
+      >
+        Load 5 First Goods
+      </button>
+
+      <button
+        className="button"
+        type="button"
+        data-cy="red-button"
+        onClick={() => handleClick(goodsAPI.getRedGoods())}
```
naziamoff, line 40: what's goodsApi?

naziamoff (changes requested):
not bad, but you should impove naming. Not really clear for now

src/App.tsx
```diff
@@ -1,27 +1,57 @@
...
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const buttonClickHandle = async (loadedGoods: Promise<Good[]>) => (
+    setGoods(await loadedGoods)
+  );
```
naziamoff, line 12: still I can't get what does this function do. Don't name your functions like `handleClickSomewhere`, because it does not give me **any** additional information. I should understand **exactly** what it does by its name.

src/App.tsx
```diff
@@ -1,27 +1,57 @@
...
+      <h1>Dynamic List of Goods</h1>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        className="button"
+        type="button"
+        data-cy="all-button"
+        onClick={() => buttonClickHandle(goodsFilterFunctionAPI.getAll())}
```
naziamoff, line 23: let's create a separate handler for each case, now it looks quite messy
```suggestion
        onClick={() => getAllTodos()}
```

naziamoff (changes requested):

src/App.tsx
Kavster92: userDefinedCallOnAPI

is that clearer as a variable name? Or do you want the variable to describe the asynchronous function calling for goods?
I'm a bit unsure what you mean here...

src/App.tsx
naziamoff: no, just get rid of `goodsFilterFunctionAPI` - it makes your code really complicated

naziamoff: oh, ok, I see now

naziamoff: this way you have it now makes much more sense

naziamoff (changes requested):
import * as goodsFilterFunctionAPI from './api/goods'; - you shouldn’t do that
Just import it function by function and use the same way

meforoma (approved):
👏🏼

PR #1273, opened 2023-02-08
Added the task solution

oleksusov:
[DEMO LINK](https://oleksusov.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,55 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRed } from './api/goods';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,55 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const allGoods = async () => {
```
polosanya, line 10:
```suggestion
  const loadAllGoods = async () => {
```
It's better to name your functions according to these rules, so it's clear that it's a function
image

polosanya (approved):
Good job!

PR #1275, opened 2023-02-08
solution

Avichez:
[DEMO LINK](https://Avichez.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,59 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleGetAllGoods = useCallback(async () => {
+    const allGoods = await getAll();
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,16 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
```
Malva37, line 13: here you modify entered data array, not sure that its correct, you can try this :
`const goods = await getAll();

  return [...goods]
    .sort((good1, good2) => good1.name.localeCompare(good2.name))
    .splice(0, 5);`

src/App.tsx
```diff
@@ -1,27 +1,59 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleGetAllGoods = useCallback(async () => {
+    const allGoods = await getAll();
```
Malva37, line 11: you didn't  predict case if getAll() throw an error, you can catch and handle it, page should work if request isn't successful

Malva37 (approved):
Nice work, just need little fix)

src/App.tsx
Avichez: Thanks, I forgot about it.

src/App.tsx
```diff
@@ -1,27 +1,77 @@
...
+  const handleGetAllGoods = useCallback(async () => {
+    try {
+      setHasRejectRespones(false);
+      const allGoods = await getAll();
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+      setGoods(allGoods);
+    } catch (error) {
```
ArturErmolenko, line 17: it will be great to show this error in console

ArturErmolenko (approved):
GJ 👍

PR #1276, opened 2023-02-09
Feat: Add goods fetching by clicking one of the buttons

mykyta-krainik:
[DEMO LINK](https://mykyta-krainik.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,48 @@
-import React from 'react';
-import './App.scss';
+import React, { useState } from 'react';
+import { getAll, get5First, getRedGoods } from './api/goods';
+
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
+        Load all goods
+      </button>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={() => get5First().then(handleDownloadedGoods)}
```
slavanosatskyi, line 32: handleDownloadedGoods function seems to be redundant, it does nothing except setting goods. You can use just setGoods

slavanosatskyi (changes requested):
Nice work, only one minor thing to fix

witflash (changes requested):
Please don't leave unresolved comments.
If you do not exactly understand what some comments mean, ask in Slack about it.

Anastasiia-Svintsova (approved):

PR #1280, opened 2023-02-13
Task done. Have a good time!)

Bogdan-Kutasevich:
[DEMO LINK](https://Bogdan-Kutasevich.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,66 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import 'bulma';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { get5First, getAll, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const getData = async (method:string) => {
+    let data;
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+    switch (method) {
+      case 'all':
+        data = await getAll();
+        setGoods(data);
+        break;
+      case 'firstFive':
+        data = await get5First();
+        setGoods(data);
+        break;
+      case 'onlyRed':
+        data = await getRedGoods();
+        setGoods(data);
+        break;
+      default:
+        throw new Error('no such data');
+    }
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  return (
+    <div className="App">
+      <h1 className="title is-1">Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        className="button is-info is-light is-outlined"
+        onClick={() => getData('all')}
+      >
+        Load all goods
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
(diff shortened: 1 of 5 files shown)


NVyrsta (approved):
Good job 👍

src/App.tsx
```diff
@@ -1,27 +1,66 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const getData = async (method:string) => {
```
Anna-Tkachenko, line 11: will be better to create separate enum for `method` if you use some limited number of values for it

src/App.tsx
```diff
@@ -1,27 +1,66 @@
...
+    let data;
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+    switch (method) {
+      case 'all':
+        data = await getAll();
+        setGoods(data);
```
Anna-Tkachenko, line 17: don't repeat yourself: you call `setGoods` after each api request
you can move this call after the `switch \ case` and call it only once

src/api/goods.ts
```diff
@@ -8,12 +8,19 @@ export function getAll(): Promise<Good[]> {
...
+    .slice(0, 5);
 };
 
-export const getRedGoods = () => {
-  return getAll()
-    .then(goods => goods); // get only red
+export const getRedGoods = async () => {
+  const result = await getAll();
+
+  return [...result]
```
Anna-Tkachenko, line 24: you shouldn't make copy here, `filter` doesn't mutate the initial array

src/App.tsx
```diff
@@ -1,27 +1,66 @@
...
+        break;
+      case 'firstFive':
+        data = await get5First();
+        setGoods(data);
+        break;
+      case 'onlyRed':
+        data = await getRedGoods();
```
Anna-Tkachenko, line 24: if there will be any issues on server, your code won't work
let's also add functionality for handling possible errors while making server request

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,72 @@
...
+          break;
+        case Select.firstFive:
+          data = await get5First();
+          break;
+        case Select.onlyRed:
+          data = await getRedGoods();
+          break;
+        default:
+          data = null;
+      }
```
Anna-Tkachenko, line 29: good job!
you can also pass needed function to make api call directly in function arguments instead of separate `method` variable

src/GoodsList.tsx
```diff
@@ -2,15 +2,28 @@ import React from 'react';
...
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
-        {good.name}
-      </li>
-    ))}
-  </ul>
+export const GoodsList: React.FC<Props> = React.memo(
+  ({ goods, dataError }) => (
+    <>
+      {dataError && <p className="error has-text-danger">Goods not foun</p>}
```
Anna-Tkachenko, line 12:
```suggestion
      {dataError && (
        <p className="error has-text-danger">Goods not found</p>
      )}
```

src/App.tsx
```diff
@@ -1,27 +1,72 @@
...
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { get5First, getAll, getRedGoods } from './api/goods';
+import { Select } from './types/Select';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[] | null>(null);
```
Anna-Tkachenko, line 10: the empty array will be better of `null` value for storing array data

Anna-Tkachenko (approved):

PR #1286, opened 2023-02-14
React dynamic list of goods solution

PolinaSukhorukova:
[DEMO LINK](https://PolinaSukhorukova.github.io/react_dynamic-list-of-goods/)


svitjojo (changes requested):
Please, check the first rule in the checklist.
And add a try-catch statement if we will not get data from the server.

ArturErmolenko (approved):
Well done 👍

PR #1287, opened 2023-02-14
List of goods implemented

svitjojo:
[DEMO LINK](https://svitjojo.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,44 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
-    </button>
+      try {
+        const selectedGoods = await f();
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+        setGoods(selectedGoods);
+      } catch {
+        setHasError(true);
```
ArturErmolenko, line 22: it will be great to understand what kind of error we have, you can log this error in console
```suggestion
      } catch (error) {
        setHasError(true);
```

ArturErmolenko (changes requested):

ArturErmolenko (approved):
Well done 🔥

PR #1292, opened 2023-02-14
solution

andrii-kurash:
[DEMO LINK](https://andrii-kurash.github.io/react_dynamic-list-of-goods/)

Good luck)😉

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,61 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState<Good[]>([]);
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,61 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const getAllGoods = async () => {
+    const goodsFromServer = await getAll();
```
denys-cheporniuk, line 11: what if getAll will return an error?
You should catch error here and show some message to user
The same for other requests

src/api/goods.ts
```diff
@@ -10,10 +10,12 @@ export function getAll(): Promise<Good[]> {
...
+    .then(goods => goods
+      .sort((a: Good, b: Good) => a.name.localeCompare(b.name)).slice(0, 5));
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods
+      .filter(good => good.color === 'red'));
 };
```
denys-cheporniuk, line 21: would be good to use async await here

denys-cheporniuk (changes requested):

PR #1293, opened 2023-02-14
add solution

Anastasiia-Khudych:
[Demo](https://anastasiia-khudych.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,20 +7,15 @@ import { Good } from './types/Good';
 
 export const App: React.FC = () => {
   const [goods, setGoods] = useState<Good[]>([]);
+  const [hasError, setHasError] = useState(false);
 
-  const loadAllGoods = () => {
-    getAll()
-      .then(goodsFromServer => setGoods(goodsFromServer));
-  };
-
-  const loadFirstFiveGoods = () => {
-    get5First()
-      .then(goodsFromServer => setGoods(goodsFromServer));
-  };
-
-  const loadRedGoods = () => {
-    getRedGoods()
-      .then(goodsFromServer => setGoods(goodsFromServer));
+  const loadGoods = (f:() => Promise<Good[]>) => {
+    f()
+      .then(goodsFromServer => {
+        setGoods(goodsFromServer);
+        setHasError(false);
+      })
+      .catch(() => setHasError(true));
   };
 
   return (
@@ -30,7 +25,7 @@ export const App: React.FC = () => {
       <button
         type="button"
         data-cy="all-button"
-        onClick={() => loadAllGoods()}
+        onClick={() => loadGoods(getAll)}
       >
         Load all goods
       </button>
@@ -38,7 +33,7 @@ export const App: React.FC = () => {
       <button
         type="button"
         data-cy="first-five-button"
-        onClick={loadFirstFiveGoods}
+        onClick={() => loadGoods(get5First)}
       >
         Load 5 first goods
       </button>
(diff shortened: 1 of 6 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,14 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods)
+    .then(goods => [...goods]
+      .sort((good1, good2) => good1.name.localeCompare(good2.name))
+      .slice(0, 5));
```
Dashaverz, line 16: it's unnecessary to use spread operator here since the sort() method can sort the original array in place. Creating a copy of the array can also have performance implications if the array is large. I think it's better to implement this in another way:
 `.then(goods => (
      goods.sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5)
    ));`

src/api/goods.ts
```diff
@@ -10,10 +10,14 @@ export function getAll(): Promise<Good[]> {
...
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods)
+    .then(goods => goods.filter(good => good.color === 'red'));
```
Dashaverz, line 22: stay dry `.then(goods => goods.filter(good => good.color === 'red'));`

src/GoodsList.tsx
```diff
@@ -8,7 +8,13 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li
+        key={good.id}
+        data-cy="good"
+        style={{
+          color: `${good.color}`,
+        }}
```
Dashaverz, line 16: `style={{ color: good.color }}` or do destructuring of 'good'

Dashaverz (changes requested):
GJ:) left some comments

src/api/goods.ts
Anastasiia-Khudych: Good point! Thank you)

src/App.tsx
```diff
@@ -1,27 +1,55 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [hasError, setHasError] = useState(false);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadGoods = (f:() => Promise<Good[]>) => {
```
Anna-Tkachenko, line 12: will be better to create separate type for inner function instead of adding it while describing the arguments

src/api/goods.ts
```diff
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods)
```
Anna-Tkachenko, line 13: this `then` is redundant here
you didn't add any logic here, just getting goods and then passing it to the next `then`

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,56 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [hasError, setHasError] = useState(false);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadGoods = (f: LoadFucntion) => {
```
Anna-Tkachenko, line 13:
```suggestion
  const loadGoods = (f: LoadFunction) => {
```

Anna-Tkachenko (approved):
Good job!

PR #1296, opened 2023-02-14
Develop

vasyliev-anton:
[DEMO LINK](https://vasyliev-anton.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/types/Good.ts
+++ b/src/types/Good.ts
@@ -1,4 +1,5 @@
 export interface Good {
+  [x: string]: any;
   id: number;
   name: string;
   color: string;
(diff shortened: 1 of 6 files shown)


src/types/Good.ts
```diff
@@ -1,4 +1,5 @@
 export interface Good {
+  [x: string]: unknown;
```
DarkMistyRoom, line 2: why do you need to use it? make it work without this line

DarkMistyRoom (changes requested):
Good job, just one comment

src/types/Good.ts
vasyliev-anton: Oh, right, I think I needed it at one point to check how the different code works and forgot to remove it, thanks :)

SanyaBratashchuk (approved):

PR #1301, opened 2023-02-15
first getting from server

OlehSharha:
[DEMO LINK](https://OlehSharha.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,64 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import 'bulma/css/bulma.min.css';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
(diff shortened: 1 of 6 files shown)


andrii-andriushchenko (approved):
Good job 👍

src/App.tsx
```diff
@@ -1,27 +1,84 @@
...
+    try {
+      setHasResponse(false);
+      const goodFromServer = await getRedGoods();
+
+      setGood(goodFromServer);
+    } catch (erorr) {
+      setHasResponse(true);
+      throw new Error('Server has no response');
+    }
+  }, []);
```
andrii-andriushchenko, line 47: Here we can use one mutual function for different function calls in `onClick`

src/App.tsx
```diff
@@ -1,27 +1,84 @@
...
+      setHasResponse(false);
+      const goodFromServer = await getAll();
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      setGood(goodFromServer);
+    } catch (erorr) {
+      setHasResponse(true);
+      throw new Error('Server has no response');
```
ArturErmolenko, line 21: it's better to log error.message , because it could be another error

ArturErmolenko (approved):

PR #1302, opened 2023-02-15
Solution

MFrolova-art:
[DEMO LINK](https://MFrolova-art.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,84 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,84 @@
...
+      const allGoods = await getAll();
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      setGoods(allGoods);
+    } catch {
+      setHasError(true);
+    }
+  }, []);
```
vladyslav78292, line 21: good job, but i think you dont need this empty array in the end, try to delete it, does it work without it?

src/GoodsList.tsx
```diff
@@ -7,9 +7,13 @@ type Props = {
 
 export const GoodsList: React.FC<Props> = ({ goods }) => (
```
ArturErmolenko, line 8: checklist point 1

ArturErmolenko (changes requested):

src/App.tsx
alexandra-protyanova: If the second argument is omitted, the value will never be memoized, and the useCallback and the useMemo doesn't do anything.
@vladyslav78292 bad idea)

alexandra-protyanova (approved):
Well done 👍

PR #1304, opened 2023-02-15
solution

Oleksandr-Maly:
[DEMO LINK](https://oleksandr-maly.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,56 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const handleAllGoodsButton = () => {
+    getAll()
+      .then(goodsFromServer => setGoods(goodsFromServer));
+  };
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleFirstFiveGoodsButton = () => {
+    get5First()
+      .then(goodsFromServer => setGoods(goodsFromServer));
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handleRedGoodsButton = () => {
+    getRedGoods()
+      .then(goodsFromServer => setGoods(goodsFromServer));
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={handleAllGoodsButton}
+      >
+        Load all goods
+      </button>
+
+      <button
+        type="button"
+        data-cy="first-five-button"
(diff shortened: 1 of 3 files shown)


max-rozzhalovets (approved):
Like your code. Good job!

src/App.tsx
```diff
@@ -1,27 +1,56 @@
...
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const handleAllGoodsButton = () => {
+    getAll()
+      .then(goodsFromServer => setGoods(goodsFromServer));
+  };
```
Anna-Tkachenko, line 13: check point 3 in checklist

src/api/goods.ts
```diff
@@ -3,17 +3,27 @@ import { Good } from '../types/Good';
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll(): Promise<Good[]> {
+export function getAll():Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
+    .then(response => {
+      if (!response.ok) {
+        return Promise.reject(new Error('Can`t load Goods'));
```
Anna-Tkachenko, line 10: if you return an error here, you need also process it while calling `getAll()` function in `App` component

Anna-Tkachenko (changes requested):
Will be awesome to add some styles to your page

src/App.tsx
```diff
@@ -1,27 +1,67 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const handleAllGoodsButton = () => {
+    getAll()
+      .then(goodsFromServer => {
+        if (goodsFromServer.length === 0) {
+          throw new Error('Goods are empty');
```
Anna-Tkachenko, line 15: each throwing an error will cause app crashing
probably we don't expect such behaviour

can we process error here somehow in other way? without throwing an error on the client

src/App.tsx
```diff
@@ -1,27 +1,67 @@
...
+        return (setGoods(goodsFromServer));
+      });
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handleFirstFiveGoodsButton = () => {
+    get5First()
+      .then(goodsFromServer => setGoods(goodsFromServer));
```
Anna-Tkachenko, line 24: check point 3 in checklist

Anna-Tkachenko: will be better to handle errors for each api call with `catch` block

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,81 @@
...
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[] | null>(null);
+  const [cathError, setCathError] = useState(false);
```
Anna-Tkachenko, line 11: `catchError` name will be good name for some function
but not for the variable that should store some boolean value, let's think more about variable name

src/App.tsx
```diff
@@ -1,27 +1,81 @@
...
+  const handleRedGoodsButton = async () => {
+    try {
+      const goodsFromServer = await getRedGoods();
+
+      setCathError(false);
+      setGoods(goodsFromServer);
+    } catch {
+      setCathError(true);
+    }
+  };
```
Anna-Tkachenko, line 43: for now you have repeated code in 3 functions
will be good to think about the way how you can optimize it

PR #1321, opened 2023-02-20
solution

Tanya-Grebenkina:
(https://Tanya-Grebenkina.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,12 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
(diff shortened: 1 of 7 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,12 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
+      return goods.sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5);
```
anstsot, line 14: it would be better to `slice` first and then add `sort`. that will help you to save the number of actions with data

anstsot (approved):
Great job, approved! you can add a small improvement

PR #1332, opened 2023-03-03
solved

Deniso4ka23:
[DEMO LINK](https://Deniso4ka23.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,20 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
+      const newGoods = [...goods];
+
+      newGoods.sort((a, b) => {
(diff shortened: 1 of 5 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,20 @@ export function getAll(): Promise<Good[]> {
...
+      const newGoods = [...goods];
+
+      newGoods.sort((a, b) => {
+        return a.name.localeCompare(b.name);
+      });
+
+      newGoods.length = 5;
+
+      return newGoods;
+    });
```
anstsot, line 23: this code could be optimized by using `.slice()` method. to change the length of array is not a good idea. also there is no need to create a copy of goods before sorting, you can work directly with `goods`

anstsot (changes requested):
Good job, let's just optimize `get5First` function

witflash (approved):
Well done 👍

PR #1338, opened 2023-03-09
did the task

vanyachyzh:
[DEMO LINK](https://vanyachyzh.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,66 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,66 @@
...
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const getAllGoods = async () => {
+    const allGoods = await goodsAPI.getAll();
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+    if (allGoods) {
```
etojeDenys, line 14: do you really need this if statement?

etojeDenys (changes requested):
add color for `li`'s
image

src/App.tsx
```diff
@@ -1,27 +1,72 @@
...
+
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={getRedGoods}
+      >
+        Load red goods
+      </button>
+
+      {goods && (<GoodsList goods={goods} />)}
```
VitaliyBondarenko1982, line 69:
```suggestion
      {!!goods.length && (<GoodsList goods={goods} />)}
```
or
```suggestion
   <GoodsList goods={goods} />
```

VitaliyBondarenko1982: goods always will be true.
check its length or not check nothing. 
empty array will not be rendered

VitaliyBondarenko1982 (approved):

PR #1339, opened 2023-03-09
solution

bohdan-gavrilyk:
https://bohdan-gavrilyk.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,46 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,46 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleClick = async (f:Promise<Good[]>) => {
```
vlad-demchuk, line 11: use a better parameter name instead of just `f`

vlad-demchuk (approved):

PR #1340, opened 2023-03-10
Solves

Vasyliy22:
[DEMO LINK](https://Vasyliy22.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,74 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goodsAll, setGoodsAll] = useState<Good[]>([]);
+  const [goods5First, setGoods5First] = useState<Good[]>([]);
+  const [goodsRed, setGoodsRed] = useState<Good[]>([]);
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,74 @@
...
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goodsAll, setGoodsAll] = useState<Good[]>([]);
+  const [goods5First, setGoods5First] = useState<Good[]>([]);
+  const [goodsRed, setGoodsRed] = useState<Good[]>([]);
```
agniya-i, line 13: it would be better to have one common state to set the current array which depends on what button you click
```suggestion
  const [good, setGood] = useState<Good[]>([]);
```

src/App.tsx
```diff
@@ -1,27 +1,74 @@
...
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleClickAll = () => {
+    getAll()
+      .then(goods => {
+        setGoodsAll(goods);
+        setGoods5First([]);
+        setGoodsRed([]);
```
agniya-i, line 20:
```suggestion
        setGoods(goods);
```

src/App.tsx
```diff
@@ -1,27 +1,74 @@
...
+        data-cy="red-button"
+        onClick={handleClickRedGoods}
+      >
+        Load red goods
+      </button>
+
+      <GoodsList
+        goods={goodsAll}
+        goods5First={goods5First}
+        goodsRed={goodsRed}
```
agniya-i, line 70:
```suggestion
        goods={goods}
```

agniya-i (changes requested):
You should set the color contained in the good object for the text label

Screenshot 2023-03-10 at 19 47 01

agniya-i (approved):
👍

PR #1342, opened 2023-03-11
solution

kuznetsova-anastasiia:
[DEMO LINK](https://kuznetsova-anastasiia.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,46 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleClick = async (promise: Promise<Good[]>) => {
+    setGoods(await promise);
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => handleClick(getAll())}
+      >
+        Load all goods
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={() => handleClick(get5First())}
+      >
+        Load 5 first goods
+      </button>
+
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={() => handleClick(getRedGoods())}
+      >
+        Load red goods
+      </button>
+
+      <GoodsList goods={goods} />
+    </div>
+  );
+};
--- a/src/GoodsList.tsx
+++ b/src/GoodsList.tsx
@@ -7,10 +7,18 @@ type Props = {
 
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
-        {good.name}
-      </li>
-    ))}
+    {goods.map(good => {
+      const { name, id, color } = good;
+
+      return (
+        <li
+          key={id}
+          data-cy="good"
+          style={{ color }}
+        >
+          {name}
+        </li>
+      );
+    })}
   </ul>
 );
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -1,19 +1,24 @@
 import { Good } from '../types/Good';
 
 // eslint-disable-next-line
-const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
+const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods.json`;
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
+    .then(response => response.json())
+    .catch(reason => new Error(reason));
 }
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods)
+    .then(goods => [...goods]
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .slice(0, 5));
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods)
+    .then(goods => goods.filter(good => good.color === 'red'));
 };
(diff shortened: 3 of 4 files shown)


src/api/goods.ts
```diff
@@ -5,15 +5,20 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
-    .then(response => response.json());
+    .then(response => response.json())
+    .catch(reason => new Error(reason));
 }
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods)
+    .then(goods => [...goods]
```
kirill-vypirovskyi, line 15: Do we need to copy an array that arrives from server?

kirill-vypirovskyi (changes requested):

kirill-vypirovskyi (approved):

src/App.tsx
```diff
@@ -1,27 +1,68 @@
...
-  </div>
-);
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
+          if (sort !== Sort.All) {
+            setSort(Sort.All);
+            handleClick(getAll());
+          }
```
illiarerun, line 33: can we create some type of map? SortType <-> Methods to create universal handler?

src/App.tsx
```diff
@@ -1,27 +1,68 @@
...
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [sort, setSort] = useState<Sort>();
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handleClick = async (promise: Promise<Good[]>) => {
```
illiarerun, line 18: don't forget abt `useCallback`

src/api/goods.ts
```diff
@@ -5,15 +5,28 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+      return response.json();
+    })
+    .catch(error => new Error(error));
 }
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods)
+    .then(goods => goods
```
illiarerun, line 23: why do you duplicate it?
```suggestion
    .then(goods => goods
```

src/api/goods.ts
```diff
@@ -5,15 +5,28 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .slice(0, 5));
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods)
+    .then(goods => goods.filter(good => good.color === 'red'));
```
illiarerun, line 31: same here

illiarerun (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,83 @@
...
+    }
+
+    switch (sortType) {
+      case Sort.All:
+        setSort(Sort.All);
+        getGoods(getAll());
+        break;
+
+      case Sort.FirstFive:
+        setSort(Sort.All);
```
sergii-nosachenko, line 50: Incorrect sort method

src/App.tsx
```diff
@@ -1,27 +1,83 @@
...
+        break;
+
+      case Sort.Red:
+        setSort(Sort.Red);
+        getGoods(getRedGoods());
+        break;
+
+      default:
+        break;
+    }
```
sergii-nosachenko, line 61:
```suggestion
    }
    
    setSort(sortType);
```

Instead of repeating it multiple times

src/App.tsx
```diff
@@ -1,27 +1,83 @@
...
+  }, []);
+
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
+
+      {Object.values(Sort).map(current => (
```
sergii-nosachenko, line 68: 🔥

src/GoodsList.tsx
```diff
@@ -5,12 +5,24 @@ type Props = {
...
-  <ul>
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
-        {good.name}
-      </li>
-    ))}
-  </ul>
+export const GoodsList: React.FC<Props> = React.memo(
+  ({ goods }) => {
+    return (
```
sergii-nosachenko, line 10: Useless `return`

sergii-nosachenko (changes requested):
Quick fixes to make it perfect

src/App.tsx
```diff
@@ -1,27 +1,82 @@
...
+enum Sort {
+  All = 'all-button',
+  FirstFive = 'first-five-button',
+  Red = 'red-button',
+}
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+const getButtonName = (sortType: Sort) => {
```
OleksiiBlezniuk, line 14: Nice idea to use switch/case, but I have suggestion how to make it a bit simpler
We can just create an object with enum values as keys and content strings as values and then just read it using enum value

```
const contentBySortType = {
  [Sort.All]: 'Load all goods',
  // content
}

const sortBy = Sort.All;

const content = contentBySortType[sortBy]; // 'Load all goods'
```

src/App.tsx
```diff
@@ -1,27 +1,82 @@
...
-  </div>
-);
+    default:
+      return '';
+  }
+};
+
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [sort, setSort] = useState<Sort>();
```
OleksiiBlezniuk, line 32: always set initial value for state

src/App.tsx
```diff
@@ -1,27 +1,82 @@
...
+    }
+
+    setSort(sortType);
+  }, []);
+
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
+
+      {Object.values(Sort).map(current => (
```
OleksiiBlezniuk, line 67: Nice one!

src/api/goods.ts
```diff
@@ -5,15 +5,26 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+    .then(response => {
+      if (!response.ok) {
+        return Promise.reject(
+          new Error(response.statusText),
+        );
+      }
+
+      return response.json();
+    })
+    .catch(error => new Error(error));
```
OleksiiBlezniuk, line 17: Is there any sense to throw an error if error is already thrown (we catching it here)
Instead of this - create state for error and show error message to users
(unfortunately users don't open console to see if there any error while loading data, they like to see information about it on UI)

src/App.tsx
```diff
@@ -1,27 +1,82 @@
...
+      return '';
+  }
+};
+
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [sort, setSort] = useState<Sort>();
+
+  const getGoods = useCallback(async (promise: Promise<Good[]>) => {
+    setGoods(await promise);
```
OleksiiBlezniuk, line 35: await, cool 👍 
but what if there will be an error?
and how about users with slow internet connection? I guess they can be a bit confused by fact that nothing is happen on the page when they click `load` button 🙂 

We as developers that working on product improving should keep users with us and should consider users needs, don't we? 😉

OleksiiBlezniuk (approved):
Excellent job 🔥

If you want to make it perfect - add some styles to the page 💅

PR #1346, opened 2023-03-14
Solution

stanislavcodes:
[DEMO LINK](https://stanislavcodes.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -27,6 +27,7 @@ export const App: React.FC = () => {
     return () => {
       if (currentGoodsType !== goodsType) {
         setIsLoading(true);
+        setIsLoadingError(false);
         getGoods()
           .then(setGoods)
           .catch(() => setIsLoadingError(true))
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -64,17 +64,18 @@ export const App: React.FC = () => {
         Load red goods
       </button>
 
-      {isLoading ? (
-        <p>Loading...</p>
+      {isLoadingError ? (
+        <p className="text">An error occurred while loading goods!</p>
       ) : (
         <>
-          {isLoadingError ? (
-            <p className="text">An error occurred while loading goods!</p>
+          {isLoading ? (
+            <p>Loading...</p>
           ) : (
             <GoodsList goods={goods} />
           )}
         </>
       )}
+
     </div>
   );
 };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,47 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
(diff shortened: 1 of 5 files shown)


timoxa2t (approved):
Good work

src/App.tsx
```diff
@@ -1,27 +1,71 @@
...
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+      {isLoadingError && (
+        <p className="text">An error occurred while loading goods!</p>
+      )}
```
OleksiiBlezniuk, line 42: Let's show it instead of previous goods for consistency

src/App.tsx
```diff
@@ -1,27 +1,71 @@
...
+    getGoods: () => Promise<Good[]>,
+    goodsType: GoodsType = GoodsType.Default,
+  ) => {
+    return () => {
+      if (currentGoodsType !== goodsType) {
+        getGoods()
+          .then(setGoods)
+          .catch(() => setIsLoadingError(true));
+        setCurrentGoodsType(goodsType);
+      }
```
OleksiiBlezniuk, line 32: But what if user have slow internet connection? Let's add loader, now it looks buggy

https://user-images.githubusercontent.com/84275496/226944731-03163102-ad0d-45a3-9ea5-81d8d7d52140.mov

src/api/goods.ts
```diff
@@ -5,15 +5,20 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
+    .then(response => response.json())
+    .then((goods) => goods);
```
OleksiiBlezniuk, line 9: Why do you need to return same array?

src/api/goods.ts
```diff
@@ -5,15 +5,20 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
 }
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then((goods) => goods
+      .sort(({ name: one }, { name: two }) => (
```
OleksiiBlezniuk, line 15: strange names

OleksiiBlezniuk (changes requested):
Don't forget to show loader for users when data is loading from API

OleksiiBlezniuk (changes requested):
Found a little bug 😢 

https://user-images.githubusercontent.com/84275496/226973726-abb746de-e5de-4ade-966e-0fd3b1826da0.mov

OleksiiBlezniuk (changes requested):
I see same issue still =( After error is faced at least once - all next requests will not reset error state

OleksiiBlezniuk (approved):
Good 👍 But would be muuuch better if you'll add few styles to it

stanislavcodes:
> Good 👍 But would be muuuch better if you'll add few styles to it

@OleksiiBlezniuk I've ddded some basic styles!

PR #1358, opened 2023-03-20
React_dynamic-list-of-goods

07Spirit:
[DEMO LINK](https://07Spirit.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,50 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [list, setList] = useState([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleClick = (request: Promise<Good[]>) => {
+    request.then((goods: any) => {
+      setList(goods);
+    });
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  return (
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,50 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [list, setList] = useState([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleClick = (request: Promise<Good[]>) => {
+    request.then((goods: any) => {
```
polosanya, line 14: Don't use any. Set the correct type for array of Goods

src/App.tsx
```diff
@@ -1,27 +1,50 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [list, setList] = useState([]);
```
polosanya, line 11: It would be great to type the state, so it's clear what exactly this array is

src/api/goods.ts
```diff
@@ -10,10 +10,16 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((
```
polosanya, line 14: Sort is a mutable method, so it would be great to sort the copy of goods

src/api/goods.ts
```diff
@@ -10,10 +10,16 @@ export function getAll(): Promise<Good[]> {
...
+        firstGood,
+        secondGood,
+      ) => firstGood.name.localeCompare(secondGood.name)))
+    .then(goods => [...goods].splice(0, 5)); // sort and get the first 5
```
polosanya, line 18: I think it's better to use slice instead of splice here

polosanya (changes requested):
Let's improve your code a bit)

IvanFesenko (approved):

PR #1359, opened 2023-03-21
solution

sofiatryasko:
[DEMO LINK](https://sofiatryasko.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,27 +1,64 @@
...
+  useEffect(() => {
+    getAll()
+      .then(setGoods);
+  }, []);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadAll = async () => {
+    const goodsFromServer = await getAll();
```
tetiana-ivanchenko, line 16: above you use then, here acync/await, choose 1 approach

tetiana-ivanchenko (changes requested):

PR #1367, opened 2023-03-24
Solution

Ihor-Zakharov:
[DEMO LINK](https://ihor-zakharov.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,93 @@
-import React from 'react';
+import React, {
+  useCallback, useEffect, useRef, useState,
+} from 'react';
+import { getFiveFirst, getAll, getRed } from './api/goods';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
 // import { getAll, get5First, getRed } from './api/goods';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
+
+    load();
+
+    return () => {
+      mounted = false;
+      mountedRef.current = false;
+    };
+  }, []);
+
+  const handleButtonAllClicked = useCallback(async (): Promise<void> => {
```
Volodymyr-Melnyk-Fron10der, line 38: useCallback is useless in those cases 
it's a complex topic so you can read more here if you want to dive deeper
https://www.freecodecamp.org/news/memoization-in-javascript-and-react/

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! 

I don't clearly understant your logic with mounted variable and mountedRef.  useEffect with an emply dependencies array is called  only once right after component was mounted, so why do use those variables? just load all goods and set it into the state

SerhiiKirik (approved):
Great 🔥

PR #1372, opened 2023-03-29
solution

Zelinskyi-Serhii:
[DEMO LINK](https://Zelinskyi-Serhii.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -102,7 +102,7 @@ export const App: React.FC = () => {
         Load red goods
       </Button>
       {error !== '' && <span>{getErrorMessage(error)}</span> }
-      {currentButton === PressedButton.None && allGoods.length === 0
+      {currentButton !== PressedButton.None && allGoods.length === 0
         ? <div className="lds-hourglass" />
         : <GoodsList goods={allGoods} />}
     </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -52,6 +52,7 @@ export const App: React.FC = () => {
       const goods = await getGoodsByButton(selectedButton);
 
       setAllGoods(goods);
+      setError('');
     } catch (e) {
       setError(getErrorMessage(e));
     }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -60,16 +60,13 @@ export const App: React.FC = () => {
   return (
     <div className="App">
       <h1>Dynamic list of Goods</h1>
+
       <Button
         type="button"
         data-cy="all-button"
         variant="outline-primary"
         className="Button"
-        onClick={() => {
-          if (currentButton !== PressedButton.AllGoods) {
-            getGoods(PressedButton.AllGoods);
-          }
-        }}
+        onClick={() => getGoods(PressedButton.AllGoods)}
       >
         Load all goods
       </Button>
@@ -79,11 +76,7 @@ export const App: React.FC = () => {
         data-cy="first-five-button"
         variant="outline-warning"
         className="Button"
-        onClick={() => {
-          if (currentButton !== PressedButton.FirstFive) {
-            getGoods(PressedButton.FirstFive);
-          }
-        }}
+        onClick={() => getGoods(PressedButton.FirstFive)}
       >
         Load 5 first goods
       </Button>
@@ -93,16 +86,18 @@ export const App: React.FC = () => {
         data-cy="red-button"
         variant="outline-danger"
         className="Button"
-        onClick={() => {
-          if (currentButton !== PressedButton.RedColor) {
-            getGoods(PressedButton.RedColor);
-          }
-        }}
+        onClick={() => getGoods(PressedButton.RedColor)}
       >
         Load red goods
       </Button>
-      {error !== '' && <span>{getErrorMessage(error)}</span> }
-      {currentButton !== PressedButton.None && allGoods.length === 0
+
+      {error
+      && (
+        <span>{getErrorMessage(error)}</span>
+      )}
+
+      {currentButton !== PressedButton.None
+      && allGoods.length === 0
         ? <div className="lds-hourglass" />
         : <GoodsList goods={allGoods} />}
     </div>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,27 +6,72 @@ import { GoodsList } from './GoodsList';
 import { getAll, get5First, getRedGoods } from './api/goods';
 import { Good } from './types/Good';
 
+enum PressedButton {
+  None,
+  AllGoods,
+  FirstFive,
+  RedColor,
+}
+
+function getErrorMessage(error: unknown) {
+  if (error instanceof Error) {
+    return error.message;
+  }
+
+  return String(error);
+}
+
+function getGoodsByButton(selectedButton: PressedButton) {
+  switch (selectedButton) {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [allGoods, setAllGoods] = useState<Good[]>([]);
+  const [isSelected, setIsSelected] = useState(false);
```
illiarerun, line 11: why all 3 buttons are saying `Loading` when I clock just on one of them?
image
Let's instead show some fancy loader

src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+        variant="outline-primary"
+        className="Button"
+        onClick={() => {
+          getAll().then(goods => setAllGoods(goods));
+          setIsSelected(true);
+        }}
+      >
+        {allGoods.length === 0 && isSelected
+          ? 'Loading...'
+          : 'Load all goods'}
```
illiarerun, line 29: So let's show only the text here

src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
-    </button>
+      <Button
+        type="button"
+        data-cy="first-five-button"
+        variant="outline-warning"
+        className="Button"
+        onClick={() => {
+          get5First().then(goods => setAllGoods(goods));
+          setIsSelected(true);
+        }}
```
illiarerun, line 40: look, you repeat yourself in all handlers, the only thing that changes is a method at a very start. Let's instead create universal handler that will accept all needed callback

illiarerun: Also let's rewrite it using `async/await` syntax

illiarerun: Also, [checklist](https://github.com/mate-academy/react_dynamic-list-of-goods/blob/master/checklist.md) 3rd item

src/api/goods.ts
```diff
@@ -1,19 +1,41 @@
...
-    .then(response => response.json());
+    .then(response => {
+      if (response.ok) {
+        return response.json();
+      }
+
+      return Promise.reject(Error);
+    })
+    .catch(error => {
+      window.console.warn(`${error} Not found`);
```
illiarerun, line 28: let's instead show it on the page

src/App.tsx
illiarerun: Also, everytime when I press the button, new request is created, even if this button was the last pressed one. 
image
Let's save last pressed button somehow and avoid reloading data if we press the same button again

illiarerun (changes requested):
Let's now fix some issues, and don't forget to use optimization practices (`useMemo, `useCallback`) when you'll rewrite your code

src/App.tsx
```diff
@@ -1,27 +1,110 @@
...
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
+      <Button
+        type="button"
+        data-cy="all-button"
+        variant="outline-primary"
+        className="Button"
+        onClick={() => {
+          if (currentButton !== PressedButton.AllGoods) {
```
illiarerun, line 69: do you really need this check here? you already have it in your `getGoods` function

src/App.tsx
```diff
@@ -1,27 +1,110 @@
...
+            getGoods(PressedButton.RedColor);
+          }
+        }}
+      >
+        Load red goods
+      </Button>
+      {error !== '' && <span>{getErrorMessage(error)}</span> }
+      {currentButton !== PressedButton.None && allGoods.length === 0
+        ? <div className="lds-hourglass" />
+        : <GoodsList goods={allGoods} />}
```
illiarerun, line 107: please, format it better

src/App.tsx
```diff
@@ -1,27 +1,110 @@
...
+        className="Button"
+        onClick={() => {
+          if (currentButton !== PressedButton.RedColor) {
+            getGoods(PressedButton.RedColor);
+          }
+        }}
+      >
+        Load red goods
+      </Button>
+      {error !== '' && <span>{getErrorMessage(error)}</span> }
```
illiarerun, line 104: you can shorten this check
```suggestion
      {error && <span>{getErrorMessage(error)}</span> }
```

src/App.tsx
```diff
@@ -1,27 +1,106 @@
...
+    if (selectedButton === currentButton) {
+      return;
+    }
+
+    setCurrentButton(selectedButton);
+    try {
+      const goods = await getGoodsByButton(selectedButton);
+
+      setAllGoods(goods);
+      setError('');
```
Oleksandr-kopaevich, line 55: it's better to reset the error when a request starts, otherwise, a customer will be confused by the old error while a new request is still in progress.

```suggestion
    try {
      setError('');
      const goods = await getGoodsByButton(selectedButton);

      setAllGoods(goods);
```

src/GoodsList.tsx
```diff
@@ -1,16 +1,29 @@
...
-        {good.name}
-      </li>
-    ))}
-  </ul>
+export const GoodsList: React.FC<Props> = React.memo(
+  ({ goods }) => (
+    <ListGroup>
+      {goods.map(good => {
+        const { id, color, name } = good;
+
```
Oleksandr-kopaevich, line 14: When you don't need to pass the whole object further (in this case variable `good`) then it's better to write like this:
```suggestion
      {goods.map(({ id, color, name }) => {
```
It's one line shorter and shows that we use only specific field and don't need the whole object

src/api/goods.ts
```diff
@@ -1,19 +1,44 @@
...
+function sortByName<T extends Good>(value: T[]): T[] {
+  return [...value].sort((prev, curr) => prev.name.localeCompare(curr.name));
+}
+
+function sliceByNumber(value: Good[], number: number): Good[] {
+  return [...value].slice(0, number);
+}
+
+function filterByColor(value: Good[], color: string): Good[] {
+  return [...value].filter(element => element.color === color);
```
Oleksandr-kopaevich, line 12: Methods slice and filter don't modify the array on which they are called, they return a new object, so the spread operator is redundant here

src/api/goods.ts
```diff
@@ -1,19 +1,44 @@
 import { Good } from '../types/Good';
 
+function sortByName<T extends Good>(value: T[]): T[] {
+  return [...value].sort((prev, curr) => prev.name.localeCompare(curr.name));
+}
+
+function sliceByNumber(value: Good[], number: number): Good[] {
```
Oleksandr-kopaevich, line 7: When variable contains an array inside it's better to put name in plural, like `values` instead of `value`

Oleksandr-kopaevich (approved):
The task is done correctly and you fixed the previous comments. Well done! Approved.

Oleksandr-kopaevich:
@Zelinskyi-Serhii Just fix this one failing check :)
[Test / run_tests (14.x) (pull_request) Failing after 2m](https://github.com/mate-academy/react_dynamic-list-of-goods/actions/runs/4608445220/jobs/8144208829?pr=1372)

PR #1376, opened 2023-04-01
Added React Dynamic List of Goods

ctk12:
[DEMO LINK](https://ctk12.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,15 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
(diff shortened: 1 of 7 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,15 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
+      return goods.sort((good1, good2) => good1.name.localeCompare(good2.name))
```
witflash, line 14: You could use arrow function as a callback for `then` block. It may looks better.

witflash (approved):
Good job! Well done 👍🏻

PR #1382, opened 2023-04-03
solution

Nordexz:


Yevhenii0536 (changes requested):
image
make a little bit style on your project.
also, when i press some button i didn't see that information was loaded, just showed list without any delay

PR #1386, opened 2023-04-04
Develop

milenadiamonds1:
[DEMO LINK](https://milenadiamonds1.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,67 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { get5First, getAll, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleGoods = async (type: string) => {
+    let goodsFromServer: React.SetStateAction<Good[]>;
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+    switch (type) {
+      case 'all':
+        goodsFromServer = await getAll();
+        break;
+      case 'five':
+        goodsFromServer = await get5First();
+        break;
+      case 'red':
+        goodsFromServer = await getRedGoods();
+        break;
+      default:
+        goodsFromServer = [];
+        break;
+    }
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+    setGoods(goodsFromServer);
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  return (
+    <div className="App">
+      <h1 className="title">
+        Dynamic list of Goods
+      </h1>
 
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [loading, setLoad] = useState({
+    all: false,
+    five: false,
+    red: false,
+  });
```
OleksiiBlezniuk, line 15: Interesting idea, but as for me it's strange that we run query before we get previous query result
I guess will be more logical to disable all buttons while something is loading

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
+      [type]: true,
+    }));
+    setloadingError(false);
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+    try {
+      switch (type) {
+        case 'all':
```
OleksiiBlezniuk, line 29: use enum for this purpose

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
+    all: false,
+    five: false,
+    red: false,
+  });
+  const [loadingError, setloadingError] = useState(false);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleGoods = async (type: string) => {
```
OleksiiBlezniuk, line 18: You may leave it as it is now, but this solution is not scalable
In case if you pass loader function as parameter to this function - you can create as many functions as you want, but now for that you need to add another case in this function to handle it

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
+    let goodsFromServer: React.SetStateAction<Good[]>;
```
OleksiiBlezniuk, line 19: strange type, why is variable with array of goods has type of SetStateAction?

src/api/goods.ts
```diff
@@ -5,15 +5,32 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
+    .then(response => {
+      if (!response.ok) {
+        return Promise.reject(
+          new Error(response.statusText),
+        );
```
OleksiiBlezniuk, line 12: Or you could just throw an error here

src/api/goods.ts
```diff
@@ -5,15 +5,32 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+    });
 }
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((firstGood, nextGood) => firstGood.name
+        .localeCompare(nextGood.name))
+      .slice(0, 5)); // sort and get the first 5
```
OleksiiBlezniuk, line 30:
```suggestion
      .slice(0, 5));
```

OleksiiBlezniuk: now such comments can be removed, your code should be understandable without explanation comments

src/App.tsx
OleksiiBlezniuk: fix it

OleksiiBlezniuk (changes requested):
Nice UI! Good job!
Left some comments, please fix it

PR #1389, opened 2023-04-04
Develop react_dynamic-list-of-goods

sergey-lipikhin:
[DEMO LINK](https://sergey-lipikhin.github.io/react_dynamic-list-of-goods/)


Ambulence (approved):
Cool design! But your GitHub test crashed(( I suggest you make one more commit and add all the libraries that you used here, such as "classnames" and other.  GitHub doesn't know about it. You have a message in the tests about it

src/components/LoadingButton/LoadingButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+type Props = {
+  text: string;
+  'data-cy': string;
+  onDataLoad: (stopButtonLoading: () => void, goodsGroupText: string) => void;
+};
+
+export const LoadingButton: React.FC<Props> = (props) => {
+  const {
+    text,
+    'data-cy': dataCy,
```
OleksiiBlezniuk, line 13: as for me - it's a bit strange to pass this prop written in kebab-case to convert it to camelCase to pass it to kebab-cased attribute
we can pass this prop camelCased to this component and avoid not needed converting

src/components/LoadingButton/LoadingButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    onDataLoad,
+  } = props;
+
+  const [isButtonLoading, setIsButtonLoading] = useState(false);
+
+  const handleDataLoad = async () => {
+    setIsButtonLoading(true);
+    onDataLoad(() => setIsButtonLoading(false), text);
+  };
```
OleksiiBlezniuk, line 22: wow, interesting

src/App.tsx
```diff
@@ -1,27 +1,91 @@
...
+    (getGoodsFromServer: () => Promise<Good[]>) => (
+      async (stopButtonLoading: () => void, goodsGroupText: string) => {
+        setIsLoading(true);
+        setLoadedGoodsGroupText(goodsGroupText);
+
+        try {
+          const goodsFromServer = await getGoodsFromServer();
+
+          setGoods(goodsFromServer);
+          setIsInitialized(true);
```
OleksiiBlezniuk, line 30: why do you need this state? it seem to be similar to `!hasLoadingError && !isLoading` result

src/components/ColorPalette/ColorPalette.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Color } from '../../types/Color';
+import { ColorItem } from '../ColorItem';
+
+const colors: Color[] = [
+  { id: 1, keyword: 'red' },
+  { id: 2, keyword: 'green' },
+  { id: 3, keyword: 'blue' },
+  // { id: 4, keyword: 'aqua' },
+  // { id: 5, keyword: 'orange' },
+];
```
OleksiiBlezniuk, line 12: Would be nice to create separate file with such constants to manage it in separate place

OleksiiBlezniuk (changes requested):
Good job! Awesome UI and cool idea to make color selectable 🔥 

Can't approve this task because there is a bug, please fix it
https://user-images.githubusercontent.com/84275496/230066058-44849b72-faf2-4424-a9ec-8729e2daabb5.mov

src/App.tsx
sergey-lipikhin: I need to use `setIsInitialized` to handle the situation where no goods are received from the server, and display a corresponding message. So I change it once when the user clicks on a button for the first time.

Without `setIsInitialized` a "_no goods_" message would be displayed even BEFORE the user has made his first request (like on the picture), which is not the desired behavior.
image

src/App.tsx
```diff
@@ -17,9 +17,32 @@ export const App: React.FC = () => {
...
+      ) {
+        return true;
+      }
+
+      prevGetGoodsFromServer.current = newGetGoodsFromServer;
+
+      return false;
+    },
+    [hasLoadingError],
+  );
```
sergii-nosachenko, line 35: Instead of having this monstrous logic, you can create an enum for button types and store `currentGoodsType` in the state. Then pass it to `handleGoodsLoad` as the second argument and check.

sergii-nosachenko (approved):

PR #1391, opened 2023-04-04
Solution

VadymTsyndra:
https://VadymTsyndra.github.io/react_dynamic-list-of-goods/


SanyaBratashchuk (changes requested):
Link is not working

src/api/goods.ts
```diff
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods.sort((
+      a: Good, b: Good,
+    ) => a.name.localeCompare(b.name)))
+    .then(goods => goods.slice(0, 5));
```
SerhiiKirik, line 16: Sort already know about that
```suggestion
    .then(goods => goods.sort((firstGood, secondGood) => (
      firstGood.name.localeCompare(secondGood.name)
    ))
    .then(goods => goods.slice(0, 5));
```

src/api/goods.ts
```diff
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods.sort((
```
SerhiiKirik, line 13: In your components you use async / await
but here you use than, why?)

SerhiiKirik (approved):
Well done 🚀

PR #1392, opened 2023-04-04
Develop

mstrilec:
[DEMO LINK](https://mstrilec.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,26 +1,79 @@
 import React, { useState } from 'react';
+import classNames from 'classnames';
+import 'bulma/css/bulma.min.css';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 import { Good } from './types/Good';
 
-import { getAll, get5First, getRedGoods } from './api/goods';
+import {
+  getAll,
+  get5First,
+  getRedGoods,
+  wait,
+} from './api/goods';
+import { LoadingError } from './components/LoadingError';
 
 export const App: React.FC = () => {
   const [goods, setGoods] = useState<Good[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [loadingFirstFive, setLoadingFirstFive] = useState(false);
+  const [loadingRed, setLoadingRed] = useState(false);
+  const [hasError, setHasError] = useState(false);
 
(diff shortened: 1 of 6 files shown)


andriibezkorovainyi (approved):
GJ

src/App.tsx
```diff
@@ -1,27 +1,109 @@
...
+import { LoadingError } from './components/LoadingError';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [loading, setLoading] = useState(false);
+  const [loadingFirstFive, setLoadingFirstFive] = useState(false);
+  const [loadingRed, setLoadingRed] = useState(false);
```
OleksiiBlezniuk, line 20: Let's disable buttons when something loading instead, it's strange that we have 3 states for loading and that we can make request while we didn't get previous result

src/App.tsx
```diff
@@ -1,27 +1,109 @@
...
+  const [hasError, setHasError] = useState(false);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleAllGoods = async () => {
+    setLoading(true);
+    setHasError(false);
+    await wait(1000);
```
OleksiiBlezniuk, line 26: Why do we need to increase waiting time, we already make real requests that take time for loading

src/App.tsx
```diff
@@ -1,27 +1,109 @@
...
+    setLoadingRed(false);
+
+    getRedGoods()
+      .then(setGoods)
+      .catch(() => {
+        setLoadingRed(false);
+        setHasError(true);
+      });
+  };
```
OleksiiBlezniuk, line 63: All these functions are similar, let's instead create one function that will handle loading and pass loading function callback to it

OleksiiBlezniuk (changes requested):
Looks like bug

https://user-images.githubusercontent.com/84275496/230079789-30c664e0-0366-4a74-8328-0a2821b9c493.mov

PR #1394, opened 2023-04-04
Main task completed

sasha-dranuta:
[DEMO LINK](https://sasha-dranuta.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+      <button
+        type="button"
+        className="button"
+        data-cy="red-button"
+        onClick={loadFilteredGoods}
+      >
+        Load red goods
+      </button>
+
+      <GoodsList goods={goods} />
```
Bogdan-Mykhailov, line 62: As Oleksii told us today we have to inform user that something changing when he press the button, or if something went wrong. Need to add loader and information about error

Bogdan-Mykhailov (changes requested):
good job. but for my opinion you need to make a little user friendly interface and errors handling

src/App.tsx
sasha-dranuta: Ye, i know, this was not my finall version of task

PR #1395, opened 2023-04-04
Develop

DenysDubas:
(https://DenysDubas.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,11 +19,6 @@ export const App: React.FC = () => {
   const handleGetGoods = useCallback(async (sortType: string) => {
     let goodsFromServer: React.SetStateAction<Good[]> = [];
 
-    // setChoosedType(prev => ({
-    //   ...prev,
-    //   [sortType]: true,
-    // }));
-
     try {
       switch (sortType) {
         case 'all':
@@ -87,8 +82,8 @@ export const App: React.FC = () => {
 
       {getError
         ? (
-          <h1 className="notification is-danger is-light">
-            Oops, something went wrong!
+          <h1 className="is-danger">
+            404 Error
           </h1>
         ) : (
           <GoodsList goods={goods} />
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,98 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
+import 'bulma/css/bulma.css';
+import classNames from 'classnames';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRed } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [isChoosedType, setIsChoosedType] = useState({
+    all: false,
+    fiveFirst: false,
+    onlyRed: false,
+  });
+  const [getError, setGetError] = useState(false);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleGetGoods = useCallback(async (sortType: string) => {
+    let goodsFromServer: React.SetStateAction<Good[]> = [];
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+    // setChoosedType(prev => ({
+    //   ...prev,
+    //   [sortType]: true,
+    // }));
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+    try {
+      switch (sortType) {
+        case 'all':
+          goodsFromServer = await getAll();
+          break;
+        case 'fiveFirst':
+          goodsFromServer = await get5First();
+          break;
+        case 'onlyRed':
+          goodsFromServer = await getRed();
+          break;
+        default:
+          goodsFromServer = [];
+          break;
+      }
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      setGoods(goodsFromServer);
+    } catch {
+      setGetError(true);
+    } finally {
+      setIsChoosedType(prev => ({
+        ...prev,
+        [sortType]: false,
+      }));
+    }
+  }, []);
+
+  return (
+    <div className="App">
+      <h1 className="h1">Dynamic list of Goods</h1>
+
+      <button
+        type="button"
+        data-cy="all-button"
+        className={classNames('button',
+          { 'is-loading': isChoosedType.all })}
+        onClick={() => handleGetGoods('all')}
+      >
+        Load all goods
+      </button>
+
+      <button
+        type="button"
+        data-cy="first-five-button"
+        className={classNames('button',
+          { 'is-loading': isChoosedType.fiveFirst })}
+        onClick={() => handleGetGoods('fiveFirst')}
+      >
+        Load 5 first goods
+      </button>
+
+      <button
+        type="button"
+        data-cy="red-button"
+        className={classNames('button',
+          { 'is-loading': isChoosedType.onlyRed })}
+        onClick={() => handleGetGoods('onlyRed')}
+      >
+        Load red goods
+      </button>
+
+      {getError
+        ? (
+          <h1 className="notification is-danger is-light">
+            Oops, something went wrong!
+          </h1>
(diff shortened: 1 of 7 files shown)


Anton-Tretiak (approved):
Good job!

src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
+    all: false,
+    fiveFirst: false,
+    onlyRed: false,
+  });
+  const [getError, setGetError] = useState(false);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleGetGoods = useCallback(async (sortType: string) => {
```
OleksiiBlezniuk, line 19: Is it sortType? As far as I see you didn't sort anything in this function

OleksiiBlezniuk, line 19:
```suggestion
  const handleGetGoods = useCallback(async (loadType: string) => {
```

src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
+    let goodsFromServer: React.SetStateAction<Good[]> = [];
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+    try {
+      switch (sortType) {
+        case 'all':
```
OleksiiBlezniuk, line 24: use enum for this purpose

src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
-      Load all goods
-    </button>
+  const handleGetGoods = useCallback(async (sortType: string) => {
+    let goodsFromServer: React.SetStateAction<Good[]> = [];
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+    try {
+      switch (sortType) {
```
OleksiiBlezniuk, line 23: You may leave it as it is now, but this solution is not scalable
In case if you pass loader function as parameter to this function - you can create as many functions as you want, but now for that you need to add another case in this function to handle it

src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [isChoosedType, setIsChoosedType] = useState({
+    all: false,
+    fiveFirst: false,
+    onlyRed: false,
+  });
+  const [getError, setGetError] = useState(false);
```
OleksiiBlezniuk, line 17: but why get? boolean constant is not a function, why its name represent action?

OleksiiBlezniuk, line 17:
```suggestion
  const [hasError, setHasError] = useState(false);
```

src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleGetGoods = useCallback(async (sortType: string) => {
+    let goodsFromServer: React.SetStateAction<Good[]> = [];
```
OleksiiBlezniuk, line 20: Why this variable has React.SetStateAction type? Isn't is just goods array?

src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
-      Load red goods
-    </button>
+      setGoods(goodsFromServer);
+    } catch {
+      setGetError(true);
+    } finally {
+      setIsChoosedType(prev => ({
+        ...prev,
+        [sortType]: false,
+      }));
```
OleksiiBlezniuk, line 45: Ok, but where did you set true for that property? 🙂

src/App.tsx
```diff
@@ -1,27 +1,93 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [isChoosedType, setIsChoosedType] = useState({
+    all: false,
+    fiveFirst: false,
+    onlyRed: false,
+  });
```
OleksiiBlezniuk, line 16: I guess it's kind of overkill, can't we just save last clicked type?

OleksiiBlezniuk (changes requested):
Found a bug

https://user-images.githubusercontent.com/84275496/230052138-4a3e1823-4e12-4e75-94bb-4fd2386c196b.mov

Also would be better to make error message red

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRed } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+enum Sort {
+  Default = '',
```
sergii-nosachenko, line 11: Use non-empty string value

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRed } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+enum Sort {
```
sergii-nosachenko, line 10: Why sort?

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
+  FiveFirst = 'fiveFirst',
+  OnlyRed = 'onlyRed',
+}
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [isChoosedType, setIsChoosedType] = useState(false);
```
sergii-nosachenko, line 19: Use enum with data type instead

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
+  const [hasError, setHasError] = useState(false);
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handleGetGoods = useCallback(async (loadType: string) => {
```
sergii-nosachenko, line 22: Use your enum for `loadType` too

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
+
+  return (
+    <div className="App">
+      <h1 className="h1">Dynamic list of Goods</h1>
+
+      <button
+        type="button"
+        data-cy="all-button"
+        className={classNames('button',
+          { 'is-loading': isChoosedType })}
```
sergii-nosachenko, line 59: Add boolean state for loading and use it

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
+        onClick={() => handleGetGoods('all')}
```
sergii-nosachenko, line 60:
```suggestion
        disabled={isLoading}
        onClick={() => handleGetGoods('all')}
```

Disable button to prevent abusing

src/App.tsx
```diff
@@ -1,27 +1,95 @@
...
+      <button
+        type="button"
+        data-cy="red-button"
+        className={classNames('button',
+          { 'is-loading': isChoosedType })}
+        onClick={() => handleGetGoods('onlyRed')}
+      >
+        Load red goods
+      </button>
+
```
sergii-nosachenko, line 84: Add loader component to show it when is loading

sergii-nosachenko (changes requested):
I have found issues with your app:
- buttons can be abused;
- you load the same data from the server again on the same button click;
- you don't remove errors;


https://user-images.githubusercontent.com/54940595/230323736-2ed41878-bec9-4e54-9bd4-dee4264140f5.mov

src/App.tsx
```diff
@@ -7,7 +7,7 @@ import { GoodsList } from './GoodsList';
 import { getAll, get5First, getRed } from './api/goods';
 import { Good } from './types/Good';
 
-enum Sort {
+enum Filter {
   Default = '',
```
sergii-nosachenko, line 11:
```suggestion
  Default = 'none',
```

src/App.tsx
```diff
@@ -16,37 +16,48 @@ enum Sort {
...
-          goodsFromServer = [];
-          break;
-      }
+  useEffect(() => {
+    async function fetchData() {
+      setIsLoading(true);
+      setHasError(false);
+
+      try {
+        let goodsFromServer: Promise<Good[]> | Good[] = [];
```
sergii-nosachenko, line 29:
```suggestion
        let goodsFromServer: Good[] = [];
```

sergii-nosachenko (approved):

PR #1405, opened 2023-04-05
Loading from API

olecsiy-gavrilyuk:
[DEMO LINK](https://olecsiy-gavrilyuk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -51,7 +51,7 @@ export const App: React.FC = () => {
               type="button"
               data-cy="all-button"
               onClick={() => handleGoods(getAll, GoodsType.All)}
-              disabled={currentGoodsType === 'all'}
+              disabled={currentGoodsType === GoodsType.All}
             >
               Load all goods
             </button>
@@ -60,7 +60,7 @@ export const App: React.FC = () => {
               type="button"
               data-cy="first-five-button"
               onClick={() => handleGoods(get5First, GoodsType.FirstFive)}
-              disabled={currentGoodsType === 'firstFive'}
+              disabled={currentGoodsType === GoodsType.FirstFive}
             >
               Load 5 first goods
             </button>
@@ -69,7 +69,7 @@ export const App: React.FC = () => {
               type="button"
               data-cy="red-button"
               onClick={() => handleGoods(getRedGoods, GoodsType.OnlyRed)}
-              disabled={currentGoodsType === 'onlyRed'}
+              disabled={currentGoodsType === GoodsType.OnlyRed}
             >
               Load red goods
             </button>


src/App.scss
```diff
@@ -1 +1,37 @@
-// styles go here
```
sergii-nosachenko, line 1: Replace the line comment with a block comment for better consistency. Use /* styles go here */ instead of // styles go here.

src/App.tsx
```diff
@@ -1,27 +1,70 @@
...
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const handleGoods = async (promiseFunc: () => Promise<Good[]>) => {
+    try {
+      setHasError(false);
+      setIsLoading(true);
+      const goodsFromApi = await getCallback(promiseFunc);
```
sergii-nosachenko, line 23:
```suggestion
      const goodsFromApi = await promiseFunc();
```

src/App.tsx
```diff
@@ -1,27 +1,70 @@
...
-      Load 5 first goods
-    </button>
+    return goodsFromApi;
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const handleGoods = async (promiseFunc: () => Promise<Good[]>) => {
+    try {
```
sergii-nosachenko, line 20: Filter pressing the same button twice to avoid loading the same data multiple times. Only in case there is no error.

src/App.tsx
```diff
@@ -1,27 +1,70 @@
...
+        onClick={() => handleGoods(getRedGoods)}
+      >
+        Load red goods
+      </button>
+
+      {isLoading
+        ? (
+          <svg viewBox="25 25 50 50">
+            <circle r="20" cy="50" cx="50" />
+          </svg>
```
sergii-nosachenko, line 64: Create a Loader component with its own style.

src/App.tsx
```diff
@@ -1,27 +1,70 @@
...
+        )
+        : <GoodsList goods={goods} hasError={hasError} />}
+
```
sergii-nosachenko, line 67:
```suggestion
```

src/GoodsList.tsx
```diff
@@ -2,15 +2,26 @@ import React from 'react';
...
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
-        {good.name}
-      </li>
-    ))}
-  </ul>
+export const GoodsList: React.FC<Props> = ({ goods, hasError }) => (
+  <>
+    { hasError
+      ? (<p>Error, server is unavailable</p>)
```
sergii-nosachenko, line 12: I think it's better to move to display an error to the App component. GoodsList must display only a list.

sergii-nosachenko (changes requested):

olecsiy-gavrilyuk:
> if user clicks on button with already loaded data - don't need to load it again

if I understand correct. In task said that i should load goods  every time when i click on button.
image

tiserett:
> > 
> 
> if I understand correct. In task said that i should load goods every time when i click on button.

You got it right, it means that you should fetch data on every click, but think of it, if you have already loaded a data, is it really neccessary to load it again?

olecsiy-gavrilyuk:
> > > 
> > 
> > 
> > if I understand correct. In task said that i should load goods every time when i click on button.
> 
> You got it right, it means that you should fetch data on every click, but think of it, if you have already loaded a data, is it really neccessary to load it again?

you mean if I click on the same button then i shouldnt load goods?

tiserett:
> > > > 
> > > 
> > > 
> > > if I understand correct. In task said that i should load goods every time when i click on button.
> > 
> > 
> > You got it right, it means that you should fetch data on every click, but think of it, if you have already loaded a data, is it really neccessary to load it again?
> 
> you mean if I click on the same button then i shouldnt load goods?

exactly, I edited my previous message a little bit, to give you a hint

olecsiy-gavrilyuk:
> > > 
> > 
> > 
> > if I understand correct. In task said that i should load goods every time when i click on button.
> 
> You got it right, it means that you should fetch data on every click, but think of it, if you have already loaded a data, is it really neccessary to load it again?
> 
> On load click, just check, if current goods type is the same as passed as prop goods type, if so - return from function, otherwise load goods

okay i get it , thanks

tiserett:
> > > > 
> > > 
> > > 
> > > if I understand correct. In task said that i should load goods every time when i click on button.
> > 
> > 
> > You got it right, it means that you should fetch data on every click, but think of it, if you have already loaded a data, is it really neccessary to load it again?
> > On load click, just check, if current goods type is the same as passed as prop goods type, if so - return from function, otherwise load goods
> 
> okay i get it , thanks

don't forget to disable button, to prevent possible user from misunderstoods

PR #1416, opened 2023-04-07
Solution by DK

DimaK123:
[DEMO LINK](https://DimaK123.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,73 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { get5First, getAll, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  type FilterParam = 'all-button' | 'first-five-button' | 'red-button';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,73 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  type FilterParam = 'all-button' | 'first-five-button' | 'red-button';
```
Volodymyr-Melnyk-Fron10der, line 11: type should be created out of component

src/App.tsx
```diff
@@ -1,27 +1,73 @@
...
-      Load 5 first goods
-    </button>
+  const setFilterParam = async (param: FilterParam) => {
+    let filteredArray: Good[];
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+    switch (param) {
+      case 'all-button':
```
Volodymyr-Melnyk-Fron10der, line 17: use enum instead of type FilterParam and you will be able to use it in cases

case FiltersEnum.All

Volodymyr-Melnyk-Fron10der (changes requested):
just a one little remark

src/App.tsx
DimaK123: Done!

tetiana-ivanchenko (changes requested):

DimaK123:
Done but what a matter ?

polosanya (approved):
Good job! It's always better to use variables/types/Enums instead of hardcored values to avoid possible typos and errors)

DimaK123:
> Good job! It's always better to use variables/types/Enums instead of hardcored values to avoid possible typos and errors)

Understand, thanks!

PR #1417, opened 2023-04-07
Develop

RuslanLiapin:
[DEMO LINK](https://ruslanliapin.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,17 +4,12 @@ import { GoodsList } from './GoodsList';
 import { Good } from './types/Good';
 
 import { getAll, get5First, getRedGoods } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
 
 export const App: React.FC = () => {
-  type StateSetter<T> = React.Dispatch<React.SetStateAction<T>>;
-
-  function usePromiseState<T>(
-    promise: Promise<T>,
-    initialState: T,
-  ): [T, StateSetter<T>] {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,75 @@
...
+
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={() => handleListType('red')}
+      >
+        Load red goods
+      </button>
+
+      <GoodsList goods={goodsToShow} />
```
TarasHoliuk, line 72: you should add colors inside list
image

src/App.tsx
```diff
@@ -1,27 +1,75 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  function usePromiseState(
```
TarasHoliuk, line 9: That's very nice you know how to create custom hooks, impressive :+1: (but it's better to do it outside a component)
Anyway, it's overcoding here. Also, you are doing 3 request fetch on the first render, but you should do it on the button click.

TarasHoliuk (changes requested):

ivanchuk-yurii (approved):

PR #1431, opened 2023-04-18
Dynamic list of goods

Sergey-Kovyazin:
[DEMO LINK](https://Sergey-Kovyazin.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -5,15 +5,28 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
+    .then(response => {
+      if (!response.ok) {
+        throw new Error(`${response.status} - ${response.statusText}`);
+      }
+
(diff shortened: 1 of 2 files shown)


src/api/goods.ts
```diff
@@ -5,15 +5,28 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
+    .then(response => {
+      if (!response.ok) {
+        throw new Error(`${response.status} - ${response.statusText}`);
+      }
+
+      if (response.headers.get('content-type')?.includes('aplication/json')) {
```
etojeDenys, line 13: There is a typo in the content type 'aplication/json'. It should be 'application/json' 🚩

etojeDenys (changes requested):
you should make these goods colorful

image
image

src/api/goods.ts
Sergey-Kovyazin: Thanks

tetiana-ivanchenko (approved):

PR #1434, opened 2023-04-20
React dynamic list of goods almost solved

sergey-vorobei:
[DEMO LINK](https://sergey-vorobei.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,65 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
-import { GoodsList } from './GoodsList';
+import { GoodsList } from './components/GoodList/GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const handleClick = async (
+    { currentTarget }: React.MouseEvent<HTMLButtonElement>,
+  ) => {
+    const { textContent } = currentTarget;
```
sTorba24, line 13: using text content as condition it's not best idea, you can use name attribute or id etc. Sure, it should be without spaces

sTorba24 (changes requested):
Well done! Lets improve it a little :)

polosanya (approved):
Good job!

PR #1437, opened 2023-04-23
dynamic-list-of-goods was done

pushkarskiyrodion:
[DEMO LINK](https://pushkarskiyrodion.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,68 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+import { getAll, get5First, getRedGoods } from './api/goods';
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,68 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[] | null>(null);
+  const [isDisabled, setIsDisabled] = useState('');
```
vlad-demchuk, line 11: if you use verb as name this state value should be of boolean type

src/App.tsx
```diff
@@ -1,27 +1,68 @@
...
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const loadAllGoods = async () => {
+    const items = await getAll();
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+    setIsDisabled('all');
```
vlad-demchuk, line 16: create enum for strings

make it safer

pushkarskiyrodion: Ok. Thanks for the advice :)

PR #1450, opened 2023-04-29
React Dynamic list of goods v.0.1 - Solution

NedobezhkinOleksandr:
https://NedobezhkinOleksandr.github.io/react_dynamic-list-of-goods/


src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadGoods = async (result: Promise<Good[]>) => {
```
agniya-i, line 11:
```suggestion
  const loadGoods = async (callback: Promise<Good[]>) => {
```

It's a callback function, so it would be better to rename

agniya-i (approved):

PR #1465, opened 2023-05-05
solution without checklist

iHateAntichrist:
[DEMO LINK](https://iHateAntichrist.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,56 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleGetAll = () => {
+    getAll().then(data => setGoods(data));
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handleGet5First = () => {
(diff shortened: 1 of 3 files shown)


Sbavdik (approved):
Good job!

src/api/goods.ts
```diff
@@ -1,7 +1,7 @@
+/* eslint-disable max-len */
```
SerhiiKirik, line 1: remove
```suggestion

```

src/api/goods.ts
```diff
@@ -10,10 +10,10 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods.sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5));
```
SerhiiKirik, line 13: you should handle errors in the `getAll` function

SerhiiKirik (changes requested):
Well done 🚀 
Let's handle errors 🧑‍💻

src/api/goods.ts
iHateAntichrist: i already handled it 
```
  return getAll()
    .then(goods => goods.sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5));
```

src/api/goods.ts
iHateAntichrist: i need it because API_URL have a long url

src/App.tsx
```diff
@@ -1,27 +1,56 @@
...
+  const handleGet5First = () => {
+    get5First().then(data => setGoods(data));
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const handleGetRedGoods = () => {
+    getRedGoods().then(data => setGoods(data));
+  };
```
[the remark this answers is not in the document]
SerhiiKirik: Do it

src/api/goods.ts
SerhiiKirik: Where you handle error?
Where you `catch` an error?

SerhiiKirik (changes requested):
Nice 🔥 
Let's improve your solution 🧑‍💻

PR #1470, opened 2023-05-07
Solution

vksazonov:
[DEMO LINK](https://vksazonov.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,54 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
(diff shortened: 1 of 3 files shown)


ymys1 (approved):
Well done!)

src/App.tsx
```diff
@@ -1,27 +1,54 @@
...
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleGetAll = useCallback(() => {
+    getAll().then(setGoods);
+  }, [goods]);
```
tiserett, line 13: why would you wrap this with useCallback? better do some researches when it's appropriately to use it

tiserett (approved):
better prevent user from loading data, which is currently loaded
it's also better to add loaders, while fetching data

PR #1483, opened 2023-05-08
Solution

Kavon-Skela:
[DEMO LINK](https://Kavon-Skela.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,50 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const visibleGoods = useMemo(() => {
+    return goods;
+  }, [JSON.stringify(goods)]);
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const getAllGoods = async () => setGoods(await getAll());
+  const getFirstFive = async () => setGoods(await get5First());
+  const getRed = async () => setGoods(await getRedGoods());
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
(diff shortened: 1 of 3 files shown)


Rage1502 (commented):
GJ! Bro!

DenysKrysanov (approved):
GJ!👍

src/api/goods.ts
```diff
@@ -5,15 +5,26 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((currGood, nextGood) => currGood.name.localeCompare(nextGood.name))
+      .slice(0, 5))
+    .catch(error => {
+      throw new Error(error);
+    });
```
jstmpelowycz, line 21: if you throw error this simple, you don't need to do it every time, you already handle it in `getAll`
otherwise, if you specify error kind (like if it occur here or there), catching it here and there makes sense
```suggestion
    .then(goods => (
      [...goods]
        .sort((a, b) => a.name.localeCompare(b.name))
        .slice(0, 5)
    ));
```

src/GoodsList.tsx
```diff
@@ -8,7 +8,11 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
```
jstmpelowycz, line 10: let's destructure it

src/App.tsx
```diff
@@ -1,27 +1,50 @@
...
-      Load all goods
-    </button>
+  const visibleGoods = useMemo(() => {
+    return goods;
+  }, [JSON.stringify(goods)]);
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const getAllGoods = async () => setGoods(await getAll());
```
jstmpelowycz, line 15:
```suggestion
  const handleGetAllGoods = async () => setGoods(await getAll());
```

src/GoodsList.tsx
```diff
@@ -8,7 +8,11 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
```
jstmpelowycz, line 8: wrapping it with `memo()` would be a good optimization

src/App.tsx
```diff
@@ -1,27 +1,50 @@
...
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const visibleGoods = useMemo(() => {
+    return goods;
+  }, [JSON.stringify(goods)]);
```
jstmpelowycz, line 13: using `useMemo()` with the same input and output values does not make sense as it will not provide any optimization or benefit in this scenario

jstmpelowycz (changes requested):

SerhiiKirik (approved):
Good job 🔥

PR #1508, opened 2023-05-15
solution

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,51 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goodsData, setGoods] = useState<{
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,51 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
```
VitaliyBondarenko1982, line 18: it is bas practise to make anonymous arrow functions in jsx.
make hanglers in component and pass reference here
check all similar cases

src/App.tsx
```diff
@@ -1,27 +1,51 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goodsData, setGoods] = useState<{
+    id: number, name: string, color: string }[]>([]);
```
VitaliyBondarenko1982, line 9: consider to make type or interface Good for better reading and maybe reuse it.

```suggestion
interface Good {
    id: number,
   name: string, 
   color: string
 }
   
export const App: React.FC = () => {
  const [goodsData, setGoods] = useState<Good[]>([]);
```

VitaliyBondarenko1982 (changes requested):

PR #1517, opened 2023-05-21
Done

Ilia991:
https://Ilia991.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,79 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+enum Sort {
+  ALL = 'all',
+  FIVEFIRST = 'fivefirst',
+  REDGOODS = 'redgoods',
+}
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [modGood, setModGood] = useState<Sort | undefined>(undefined);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  useEffect(() => {
+    const funcGoodsInfo = async () => {
+      let result;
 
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
+          result = await getAll().then(allGoods => setGoods(allGoods));
+          break;
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+        case Sort.FIVEFIRST:
+          result = await get5First().then(items => setGoods(items
+            .sort((item1, item2) => item1.name.localeCompare(item2.name))
+            .slice(0, 5)));
```
sTorba24, line 29: would be good to move all related logic to `get5First()`. And just result with `setGoods`

src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
+          break;
 
-    <GoodsList goods={[]} />
-  </div>
-);
+        case Sort.REDGOODS:
+          result = await getRedGoods().then(goodsInfo => setGoods(
+            goodsInfo.filter(good => good.color === 'red'),
+          ));
```
sTorba24, line 35: same here

src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [modGood, setModGood] = useState<Sort | undefined>(undefined);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  useEffect(() => {
+    const funcGoodsInfo = async () => {
+      let result;
```
sTorba24, line 19: i think it should be not in useEffect. Just change goods onClick.

src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
+  FIVEFIRST = 'fivefirst',
+  REDGOODS = 'redgoods',
+}
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [modGood, setModGood] = useState<Sort | undefined>(undefined);
```
sTorba24, line 15: use more meaningful name. Also seems you don't need it.
main point store goods depends from sortType. Sort type changed? Load new goods in `goods` state

sTorba24 (changes requested):
Good work! lets improve it
Also it should have red color
image

OleziO (approved):
Add color style to items.

PR #1520, opened 2023-05-22
solution

ghostargg:
[DEMO LINK](https://ghostargg.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,65 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+import { Good } from './types/Good';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[] | []>([]);
+  const [error, setError] = useState<null | string>(null);
```
d-wasilewski, line 12: Good job with this, also consider a place to remove the error after some action so that is doesn't persist on the website
```suggestion
  const [error, setError] = useState('');
```

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1521, opened 2023-05-23
Solution

Zibi95:
[DEMO LINK](https://Zibi95.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,54 @@
-import React from 'react';
+import React, { useReducer } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { get5First, getAll, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+type Action = {
+  payload: Good[]
+};
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+const reducer = (state: Good[], action: Action) => {
+  if (action.payload) {
+    return action.payload;
+  }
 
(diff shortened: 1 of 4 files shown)


src/GoodsList.tsx
```diff
@@ -8,7 +8,11 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
```
Radoslaw-Czerniawski, line 10: You could destructure your goods here:
```js
```

src/App.tsx
```diff
@@ -1,27 +1,54 @@
...
-    <h1>Dynamic list of Goods</h1>
+type Action = {
+  payload: Good[]
+};
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+const reducer = (state: Good[], action: Action) => {
+  if (action.payload) {
```
Radoslaw-Czerniawski, line 13: In this case it doesn't matter that much, but if would you have more actions it's better to use switch. 
A few notes: 
- in typical reducer usage both in react alone and react with redux state, you want to handle different operation on your state like resetting, deleting, adding etc. based on type of action
- if you had a couple of ifs, js interpreter would go from top to bottom through your code until it would find a satisfying result, in switch statement it immediately jumps to right case. For example we have a structure like below and want to dispatch an action of type 'setBooleansOpposite'


image

### in this reducer we will first check every if before reaching 'setBooleansOpposite'
image

### in this reducer we will immediately jump into 'setBooleanOpposite'
image

Radoslaw-Czerniawski: Generally reducers are more suitable for handling complex states, in this task a simple useState would suffice,  however remember for future reference similar pattern if you have to handle something more complex

Radoslaw-Czerniawski (approved):
Left some comments on using `useReducer` hook. Approving this pr as it looks nice. :)
Remember to destructure your arguments if possible, and use adequate solutions for your problems.

PR #1522, opened 2023-05-23
solution v1

bjeziorek:
[DEMO LINK](https://bjeziorek.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,6 +7,7 @@ import { Good } from './types/Good';
 
 export const App: React.FC = () => {
   const [goods, setGoods] = useState<Good[]>([]);
+  const [error, setError] = useState('');
 
   const loadAll = async () => {
     const allGoods = await goodsAPI.getAll();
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,27 +1,87 @@
...
+        data-cy="all-button"
+        onClick={() => setMode(Modes.all)}
+      >
+        Load all goods
+      </button>
+
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={() => setMode(Modes.just5)}
```
d-wasilewski, line 71: I think it would be more concise and simpler if you just used the methods here. There is not any additional logic in switch cases, and doing these things with modes just creates some unnecessary in this case code. You can directly use these functions as you have them all defined anyway
```suggestion
        onClick={loadFirst5Goods}
```

src/App.tsx
```diff
@@ -1,27 +1,87 @@
...
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [mode, setMode] = useState<Modes>(Modes.none);
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  useEffect(() => {
```
d-wasilewski, line 19: I don't think you need `useEffect` in this task

d-wasilewski (changes requested):

kasiakubisiak (approved):

PR #1524, opened 2023-05-23
solution

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,66 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,66 @@
...
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const getFive = async () => {
+    try {
+      setGoods(await get5First());
+    } catch {
+      throw new Error('Not found');
+    }
+  };
```
Viktor-Kost, line 25: 1) By separating the assignment and state update, the code becomes more readable and allows for potential future modifications or additional error handling.
2)* as recommendation: create state for an error to set an error.message there and then display it to the user or temporary display this error to console (console.error).

I'd recommend to use like this:
```suggestion
const getFive = async () => {
  try {
    const goods = await get5First();
    setGoods(goods);
  } catch {
    throw new Error('Not found');
  }
};
```

Viktor-Kost (approved):
GJ!
Left a recommendation to improve your solution

PR #1535, opened 2023-05-27
message

EvgenyNikitinEvgeny:
https://EvgenyNikitinEvgeny.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,61 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [allgoods, setAllGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const getGoodsAll = async () => {
+    const goods = await getAll();
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+    setAllGoods(goods);
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const getGet5First = async () => {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,61 @@
...
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [allgoods, setAllGoods] = useState<Good[]>([]);
```
VitaliyBodnar, line 8: you need to use camelCase here 
```suggestion
  const [allGoods, setAllGoods] = useState<Good[]>([]);
```

src/App.tsx
```diff
@@ -1,27 +1,61 @@
...
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+    setAllGoods(goods);
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const getGet5First = async () => {
```
VitaliyBodnar, line 16: not correct name
 - 'get' is duplicated
 - It's not good to use a number in const name

 You can rename it to getFirstFiveGoods

src/App.tsx
```diff
@@ -1,27 +1,61 @@
...
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+    setAllGoods(goods);
+  };
 
-    <GoodsList goods={[]} />
-  </div>
-);
+  const getGetRedGoods = async () => {
```
VitaliyBodnar, line 22: 'get' is duplicated in the const name

src/api/goods.ts
```diff
@@ -10,10 +10,11 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods.sort((a, b) => a.name.localeCompare(b.name))
+      .slice(0, 5));
```
VitaliyBodnar, line 14: this slice() should be inside .then()

VitaliyBodnar (changes requested):

loralevitska (approved):
Well done)

PR #1538, opened 2023-05-29
Solution

NadieinOleh:
https://NadieinOleh.github.io/react_dynamic-list-of-goods/


demetrxx (approved):
Generally, well done 👍
Take into account remarks bellow, though.

PR #1545, opened 2023-06-01
Develop

Vaskovics:
<a href="https://vaskovics.github.io/react_dynamic-list-of-goods/">Demo</a>


src/App.tsx
```diff
@@ -1,27 +1,42 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => getAll().then(setGoods)}
```
loralevitska, line 18: Better move such functions to separate function and add catch case to avoid breaking the page when the button is clicked

PR #1563, opened 2023-06-25
added work with API for goodsList component

AlexKherson:
[DEMO LINK](https://alexKherson.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,59 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import { Button } from '@mui/material';
 import { GoodsList } from './GoodsList';
+import { get5First, getAll, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good []>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const loadAllGoods = async () => {
+    await getAll()
+      .then(responce => setGoods(responce));
+  };
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const load5First = async () => {
+    await get5First()
+      .then(responce => setGoods(responce));
+  };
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,59 @@
...
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good []>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const loadAllGoods = async () => {
+    await getAll()
+      .then(responce => setGoods(responce));
```
dinaminahi, line 13:
```suggestion
   const allGoods = await getAll();
   setGoods(allGoods);
```

src/App.tsx
```diff
@@ -1,27 +1,59 @@
...
+  };
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const load5First = async () => {
+    await get5First()
```
dinaminahi, line 17: same here

src/App.tsx
```diff
@@ -1,27 +1,59 @@
...
+      .then(responce => setGoods(responce));
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const loadRedGoods = async () => {
+    await getRedGoods()
```
dinaminahi, line 22: and here

dinaminahi (changes requested):
Good job 👍 
Let's fix one issue:
`getAll()` returns a promise, and you want to wait for it to resolve before using the response to set the goods variable. To achieve this, you can either use await or then(), but not both together.

src/GoodsList.tsx
```diff
@@ -1,16 +1,21 @@
...
 
 export const GoodsList: React.FC<Props> = ({ goods }) => (
-  <ul>
+  <List sx={{ width: '100%', bgcolor: 'background.paper' }}>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <ListItem
+        key={good.id}
+        data-cy="good"
+        style={{ color: `${good.color}` }}
```
33Orange, line 15: Let's create class names for colors and add class names from `good.color`

33Orange (changes requested):
Good job

A few moments to improve

33Orange (approved):
Well done

PR #1564, opened 2023-06-26
add task solution

EugeneSherepa:
[DEMO LINK](https://EugeneSherepa.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((good1, good2) => good1.name.localeCompare(good2.name))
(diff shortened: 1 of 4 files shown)


dinaminahi (approved):
Well done 💪

src/api/goods.ts
```diff
@@ -10,10 +10,13 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((good1, good2) => good1.name.localeCompare(good2.name))
```
dffUqp, line 14: bad practice to use a number in variable naming, I think we can use `a` and `b` in sort method. 
```suggestion
      .sort((a, b) => a.name.localeCompare(b.name))
```

dffUqp (approved):
LGTM:fire:

PR #1566, opened 2023-06-26
Solution

amashutikov:
[DEMO LINK](https://amashutikov.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -5,15 +5,26 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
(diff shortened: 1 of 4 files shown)


dffUqp (approved):

src/api/goods.ts
```diff
@@ -5,15 +5,26 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+    });
 }
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((good1, good2) => (
+        good1.name.localeCompare(good2.name)
+      ))
```
dffUqp, line 22: bad practice to use a number in variable naming, I think we can use `a` and `b` in sort method.

PR #1569, opened 2023-06-27
Solution

Shevtsooov:
[DEMO LINK](https://Shevtsooov.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,27 +1,55 @@
...
+      throw new Error();
+    }
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handleGetAll = useCallback(() => loadGoods(getAll), []);
+  const handleGet5First = useCallback(() => loadGoods(get5First), []);
+  const handleGetRedGoods = useCallback(() => loadGoods(getRedGoods), []);
```
dffUqp, line 22: I think we need to remove `useCallback` here, because we prevent render for 0 components

src/App.tsx
```diff
@@ -1,27 +1,55 @@
...
+export const App: React.FC = () => {
+  const [filteredGoods, setFilteredGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadGoods = (callback: () => Promise<Good[]>) => {
+    try {
+      callback()
+        .then(goods => setFilteredGoods(goods));
```
dffUqp, line 14: let's rewrite this with async/await

dffUqp (changes requested):

Shevtsooov:
@dffUqp Thanks for the review, the changes have been applied)

dffUqp (approved):
🚀

PR #1570, opened 2023-06-27
Solution

skinnyjazz:
[DEMO LINK](https://skinnyjazz.github.io/react_dynamic-list-of-goods/)

I added additional functionality here to avoid unnecessary data reloading when it's not needed.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,64 @@
-import React from 'react';
+import { FC, useRef, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, getSomeGoods, getGoodsSortedByColor } from './api/goods';
+import { Good, QueryType } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const quaryType = useRef<QueryType | null>(null);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const getAllGoods = async () => {
+    if (quaryType.current !== QueryType.ALL) {
+      setGoods(await getAll());
+      quaryType.current = QueryType.ALL;
+    }
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const getFiveGoods = async () => {
+    if (quaryType.current !== QueryType.SOME) {
+      setGoods(await getSomeGoods(5));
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,64 @@
...
+export const App: FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const quaryType = useRef<QueryType | null>(null);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const getAllGoods = async () => {
+    if (quaryType.current !== QueryType.ALL) {
+      setGoods(await getAll());
```
dffUqp, line 14: we can get error from our API, let's handle this case

src/App.tsx
```diff
@@ -1,27 +1,64 @@
...
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const getFiveGoods = async () => {
+    if (quaryType.current !== QueryType.SOME) {
+      setGoods(await getSomeGoods(5));
+      quaryType.current = QueryType.SOME;
+    }
+  };
```
dffUqp, line 24: I think if we invert condition it will improve code readability(optional change)

```suggestion
  const getFiveGoods = async () => {
    if (quaryType.current === QueryType.SOME) {
      return;
    }
    
    setGoods(await getSomeGoods(5));
    quaryType.current = QueryType.SOME;
  };
```

src/api/goods.ts
```diff
@@ -8,12 +8,15 @@ export function getAll(): Promise<Good[]> {
...
 }
 
-export const get5First = () => {
+export const getSomeGoods = (numOfGoods: number): Promise<Good[]> => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods.sort(
+      (good, nextGood) => good.name.localeCompare(nextGood.name),
+    ))
+    .then(goods => goods.splice(0, numOfGoods));
```
dffUqp, line 16: can we use just method cahining in this case and not call extra `.then()`?

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

src/App.tsx
```diff
@@ -1,27 +1,82 @@
...
-// import * as goodsAPI from './api/goods';
+import { getAll, getSomeGoods, getGoodsSortedByColor } from './api/goods';
+import { Good, QueryType } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const quaryType = useRef<QueryType | null>(null);
```
33Orange, line 10:
```suggestion
  const queryType = useRef<QueryType | null>(null);
```

33Orange (approved):
Good job

PR #1572, opened 2023-06-27
solution

serentinos:
[DEMO LINK](https://serentinos.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,54 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { get5First, getAll, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  async function handleLoadAll() {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const onClickHandle = async (
+    f: () => Promise<Good[]>,
+  ) => {
+    setGoods(await f());
+  };
```
33Orange, line 15: Create 3 separate handlers for buttons

```suggestion
  const handleClickAllGoods = (event: ...) => {
    event.preventDefault();
    setGoods(getAll());
  };
```

src/App.tsx
```diff
@@ -1,27 +1,48 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => onClickHandle(getAll)}
```
33Orange, line 24:
```suggestion
        onClick={handleClickAllGoods}
```

src/api/goods.ts
```diff
@@ -3,17 +3,27 @@ import { Good } from '../types/Good';
...
+  } catch (error) {
+    throw new Error('Cant Load data');
+  }
 }
 
-export const get5First = () => {
-  return getAll()
-    .then(goods => goods); // sort and get the first 5
+export const get5First = async () => {
+  const response = await getAll();
```
dffUqp, line 18: What will we do if `getAll` return an error?

dffUqp (approved):
GJ✔️

one thing to improve the code is to disable the button when data is loading to avoid extra requests to API

PR #1573, opened 2023-06-27
made react task dynamic-list-of-goods

vladyslav-lapshyn:
[DEMO LINK](https://vladyslav-lapshyn.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,118 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import {
+  Container,
+  Stack,
+  Button,
+  Typography,
+} from '@mui/material';
+
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [
+    allButtonVariant,
+    setAllButtonVariant,
+  ] = useState<'outlined' | 'contained'>('outlined');
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const [
+    firstFiveButtonVariant,
+    setFirstFiveButtonVariant,
+  ] = useState<'outlined' | 'contained'>('outlined');
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const [
+    redButtonVariant,
+    setRedButtonVariant,
+  ] = useState<'outlined' | 'contained'>('outlined');
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const handleLoadAll = async () => {
+    if (allButtonVariant === 'outlined') {
+      setGoods(await getAll());
+      setAllButtonVariant('contained');
+      setFirstFiveButtonVariant('outlined');
+      setRedButtonVariant('outlined');
+    } else {
+      setGoods([]);
+      setAllButtonVariant('outlined');
+    }
+  };
 
-    <GoodsList goods={[]} />
-  </div>
-);
+  const handle5First = async () => {
+    if (firstFiveButtonVariant === 'outlined') {
+      setGoods(await get5First());
+      setAllButtonVariant('outlined');
+      setFirstFiveButtonVariant('contained');
+      setRedButtonVariant('outlined');
+    } else {
+      setGoods([]);
+      setFirstFiveButtonVariant('outlined');
+    }
+  };
+
+  const handleRed = async () => {
+    if (redButtonVariant === 'outlined') {
+      setGoods(await getRedGoods());
+      setAllButtonVariant('outlined');
+      setFirstFiveButtonVariant('outlined');
+      setRedButtonVariant('contained');
+    } else {
+      setGoods([]);
+      setRedButtonVariant('outlined');
+    }
+  };
+
+  return (
+    <div className="App">
+      <Container maxWidth="sm">
+
+        <Typography
+          variant="h3"
+          component="h1"
+          align="center"
+          sx={{
+            margin: '30px 0',
+          }}
+        >
+          Dynamic list of Goods
+        </Typography>
+
+        <Stack
+          spacing={2}
+          direction="row"
+          sx={{
+            justifyContent: 'center',
+            marginBottom: '30px',
+          }}
+        >
+          <Button
+            variant={allButtonVariant}
+            type="button"
+            onClick={handleLoadAll}
+          >
+            Load all goods
+          </Button>
+          <Button
+            variant={firstFiveButtonVariant}
+            type="button"
+            onClick={handle5First}
+          >
+            Load 5 first goods
+          </Button>
+          <Button
+            variant={redButtonVariant}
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,118 @@
...
+  const handleLoadAll = async () => {
+    if (allButtonVariant === 'outlined') {
+      setGoods(await getAll());
+      setAllButtonVariant('contained');
+      setFirstFiveButtonVariant('outlined');
+      setRedButtonVariant('outlined');
+    } else {
+      setGoods([]);
+      setAllButtonVariant('outlined');
+    }
```
dffUqp, line 41: I think if we invert condition it will improve code readability(optional change)
```suggestion
     if (allButtonVariant !== 'outlined') {
       setGoods([]);
       setAllButtonVariant('outlined');
       return
     } 
    
      setGoods(await getAll());
      setAllButtonVariant('contained');
      setFirstFiveButtonVariant('outlined');
      setRedButtonVariant('outlined');
```

src/App.tsx
```diff
@@ -1,27 +1,118 @@
...
+
+  const [
+    firstFiveButtonVariant,
+    setFirstFiveButtonVariant,
+  ] = useState<'outlined' | 'contained'>('outlined');
+
+  const [
+    redButtonVariant,
+    setRedButtonVariant,
+  ] = useState<'outlined' | 'contained'>('outlined');
```
dffUqp, line 30: 1. Can we create one useState which will contain object?
2. I think we need to create some const or enum for `outlined` and `contained`

src/App.tsx
```diff
@@ -1,27 +1,118 @@
...
+      setRedButtonVariant('outlined');
+    } else {
+      setGoods([]);
+      setAllButtonVariant('outlined');
+    }
+  };
+
+  const handle5First = async () => {
+    if (firstFiveButtonVariant === 'outlined') {
+      setGoods(await get5First());
```
dffUqp, line 46: What will we do if `get5First` return an error?

dffUqp (changes requested):

src/App.tsx
vladyslav-lapshyn: @dffUqp it's awesome, thanks

src/App.tsx
vladyslav-lapshyn: @dffUqp I hope I understand correctly, I added `catch` to two functions `get5First` and `getRedGoods` in file `goods.ts`

src/App.tsx
```diff
@@ -1,27 +1,139 @@
...
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
+
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+
+  enum ButtonVariant {
+    outlined = 'outlined',
+    contained = 'contained',
+  }
```
dffUqp, line 21: can we create a separate file for this enum?

src/App.tsx
```diff
@@ -1,27 +1,139 @@
...
+
+  const [buttonVariants, setButtonVariants] = useState({
+    all: ButtonVariant.outlined,
+    firstFive: ButtonVariant.outlined,
+    red: ButtonVariant.outlined,
```
dffUqp, line 26: I think it would be really cool if we create an initial state variable and then reuse it on lines 53, 33, and 73 of your code.

src/App.tsx
```diff
@@ -1,27 +1,139 @@
...
+
+      return;
+    }
+
+    setGoods(await getAll());
+    setButtonVariants({
+      all: ButtonVariant.contained,
+      firstFive: ButtonVariant.outlined,
+      red: ButtonVariant.outlined,
+    });
```
dffUqp, line 46:
```suggestion
    setButtonVariants((prevState) => ({
      ...prevState,
      all: ButtonVariant.contained,
    }));
```

src/api/goods.ts
```diff
@@ -5,15 +5,32 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
+    .then(response => {
+      if (!response.ok) {
+        throw new Error('Cannot fetch goods');
+      }
```
dffUqp, line 11: you return error in `getAll` then in `handleRed` and `handle5First` try to put this error in setState variable

here is my suggestion how to prevent this behavior:
```suggestion
    if (!response.ok) {
      console.error("Cannot fetch goods");
      return [];
    }
```

dffUqp (changes requested):
:smiley:

src/App.tsx
vladyslav-lapshyn: @dffUqp I didn't change solution in that situation, if I use your advice I can press all 3 buttons at the same time (of course in list I see last result, but all 3 buttons can be in `contained` state in at the same time) 

image

dffUqp: oh, sorry😢

try this 
```
    setButtonVariants({
       ...initialButtonVariants,
      all: ButtonVariant.contained
    });
```

src/api/goods.ts
```diff
@@ -5,15 +5,41 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+      const first5Goods = sortedGoods.slice(0, 5);
+
+      return first5Goods;
+    })
+    .catch(error => {
+      // eslint-disable-next-line no-console
+      console.error('Cannot fetch 5 first goods', error);
+
+      return [];
+    });
```
dffUqp, line 33: when you return a initial value(empty array) in getAll you can remove this code from `get5First` and `getRedGoods`

dffUqp (changes requested):
sorry for a lot of comments🥲

src/api/goods.ts
vladyslav-lapshyn: @dffUqp thanks, it's important experience for me

src/App.tsx
vladyslav-lapshyn: @dffUqp WOW, it looks awesome, honestly I didn't think of that at all. Thank you very much

src/api/goods.ts
```diff
@@ -5,15 +5,37 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+    .then(goods => {
+      const sortedGoods = goods.sort((a, b) => a.name.localeCompare(b.name));
+      const first5Goods = sortedGoods.slice(0, 5);
+
+      return first5Goods;
+    })
+    .catch(error => {
+      // eslint-disable-next-line no-console
+      console.error('Cannot fetch 5 first goods', error);
+    });
```
dffUqp, line 31:
```suggestion

```

src/api/goods.ts
```diff
@@ -5,15 +5,37 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods.filter(good => good.color === 'red'))
+    .catch(error => {
+      // eslint-disable-next-line no-console
+      console.error('Cannot fetch red goods', error);
+    });
```
dffUqp, line 40:
```suggestion

```

dffUqp (approved):
Perfect🚀

why did I suggest you return an empty array from `getAll()`?
In lines 33, 48 and 63 of your code you put "promise" directly in the state, this is a bad case because we can get an error from the API and set this error in our state, so we need to handle the error in your case you needed to wrap your code to `try/catch`, but the shorter solution is to return [] from our API if an error applies.

PR #1576, opened 2023-06-27
solution

vitvadym:
[DEMO LINK](https://vitvadym.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -44,6 +44,7 @@ export const App: React.FC = () => {
       <h1>Dynamic list of Goods</h1>
 
       <Button
+        type="button"
         variant="contained"
         data-cy="all-button"
         onClick={onAddAll}
@@ -52,6 +53,7 @@ export const App: React.FC = () => {
       </Button>
 
       <Button
+        type="button"
         variant="contained"
         data-cy="first-five-button"
         onClick={onAddFive}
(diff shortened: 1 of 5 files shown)


src/api/goods.ts
```diff
@@ -17,3 +17,11 @@ export const getRedGoods = () => {
   return getAll()
     .then(goods => goods); // get only red
 };
+
+export const sortByName = (data: Good[]) => {
+  return data.sort((a: Good, b:Good) => a.name.localeCompare(b.name));
+};
```
33Orange, line 23: Move this logic to `get5First`

src/api/goods.ts
```diff
@@ -17,3 +17,11 @@ export const getRedGoods = () => {
...
+
+export const getRedColor = (data: Good[]) => {
+  return data.filter(elem => elem.color === 'red');
+};
```
33Orange, line 27: Move this logic to `getRedGoods`

src/App.tsx
```diff
@@ -1,27 +1,84 @@
...
-      Load red goods
-    </button>
+  const onAddFive = async () => {
+    const data = await get5First();
+    const sortedData = sortByName(data);
 
-    <GoodsList goods={[]} />
-  </div>
-);
+    setGoods(sortedData.slice(0, 5));
```
33Orange, line 28: Move slice to `get5First()` function and return only array which you will set to goods state

33Orange (changes requested):
Almost done

33Orange (approved):
Good job

PR #1581, opened 2023-06-27
add dinamic list of goods solution

opiquem:
[DEMO LINK](https://maksym-bobryk.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/GoodsList.tsx
+++ b/src/GoodsList.tsx
@@ -7,16 +7,16 @@ type Props = {
   goods: Good[]
 };
 
-export const GoodsList: React.FC<Props> = React.memo(({ goods }) => {
-  const classes = (good: Good) => (
-    cn(
-      {
-        'has-text-link': good.color === 'blue',
-        'has-text-danger': good.color === 'red',
-        'has-text-success': good.color === 'green',
-      },
-    ));
+const getClasses = (good: Good) => (
+  cn(
+    {
+      'has-text-link': good.color === 'blue',
+      'has-text-danger': good.color === 'red',
+      'has-text-success': good.color === 'green',
+    },
+  ));
 
+export const GoodsList: React.FC<Props> = React.memo(({ goods }) => {
   return (
     <div className="box table-container">
       <table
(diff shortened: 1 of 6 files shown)


src/api/goods.ts
```diff
@@ -5,15 +5,24 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+    .then(goods => goods
+      .sort((g1, g2) => g1.name.localeCompare(g2.name))
+      .slice(0, 5));
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods
+      .filter(good => good.color === 'red'));
```
33Orange, line 27:
```suggestion
    .then(goods => goods.filter(good => good.color === 'red'));
```

src/GoodsList.tsx
```diff
@@ -1,16 +1,67 @@
...
-  <ul>
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
-        {good.name}
-      </li>
-    ))}
-  </ul>
-);
+export const GoodsList: React.FC<Props> = React.memo(({ goods }) => {
+  const classes = (good: Good) => (
```
33Orange, line 11:
```suggestion
  const getClasses = (good: Good) => (
```

src/GoodsList.tsx
```diff
@@ -1,16 +1,67 @@
...
+    cn(
+      {
+        'has-text-link': good.color === 'blue',
+        'has-text-danger': good.color === 'red',
+        'has-text-success': good.color === 'green',
+      },
+    ));
```
33Orange, line 18: Move this function outside your component to avoid re-creating every render.

33Orange (changes requested):
Good job.

A few moments to improve

src/api/goods.ts
```diff
@@ -5,15 +5,23 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+
+      return response.json();
+    });
 }
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((g1, g2) => g1.name.localeCompare(g2.name))
```
dffUqp, line 20: bad practice to use a number in variable naming, I think we can use `a` and `b` in sort method.

dffUqp (approved):
GJ✔️

PR #1582, opened 2023-06-28
solution

kirilshorin:
[DEMO LINK](https://kirilshorin.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/GoodsList.tsx
+++ b/src/GoodsList.tsx
@@ -7,8 +7,12 @@ type Props = {
 
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
+    {goods.map((good) => (
+      <li
+        key={good.id}
+        data-cy="good"
+        style={{ color: `${good.color}` }}
+      >
         {good.name}
       </li>
     ))}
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -10,10 +10,11 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods.sort((good1, good2) => (
+      good1.name.localeCompare(good2.name))).slice(0, 5)); // sort and get the first 5
```
dffUqp, line 14: bad practice to use a number in variable naming, I think we can use a and b in sort method.

src/GoodsList.tsx
```diff
@@ -7,8 +7,12 @@ type Props = {
 
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
+    {goods.map((good) => (
+      <li
+        key={good.id}
+        data-cy="good"
+        style={{ color: `${good.color}` }}
```
dffUqp, line 14: better use classnames here and add some styles to this classes
```suggestion
        className={good.color}
```

PR #1590, opened 2023-06-29
Solved

olehmahera:
[DEMO LINK](https://olehmahera.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,59 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { getAll, get5First, getRed } from './api/goods';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleClickGetAll = async () => {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,59 @@
...
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleClickGetAll = async () => {
+    const allGoods = await getAll();
```
YeMyr, line 11: Hi! It would be better to create one common handler for all onClick, and pass functions of request into the parametr as callbacks:
`
const commonHandler = (callback) => {
  const requestedGoods = await callback();
  setGoods(requestedGoods);
};
...
onClick = {commonHandler(getAll())}
onClick = {commonHandler(get5First())}
onClick = {commonHandler(getRedGoods())}
`

src/GoodsList.tsx
```diff
@@ -8,7 +8,11 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li
+        key={good.id}
+        data-cy="good"
+        className={good.color}
```
YeMyr, line 14: 👍 nice, solution,
I'll grab it for me

Mykola-Maik (approved):
Good job! I have the same solution with a separate handleClick. I don't know how it will be better, but it looks nice))

33Orange (approved):
Good job

PR #1592, opened 2023-06-29
add task solution

KatyaZaremska:
[DEMO LINK](https://KatyaZaremska.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,53 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [product, setProduct] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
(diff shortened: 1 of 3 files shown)


YeMyr (approved):

src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [product, setProduct] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const loadAllGoods = async () => {
```
33Orange, line 11: Let's use `handle` or `on` when creating an event function.

```suggestion
  const onClickAllGoods = async () => {
```

src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+    setProduct(goods);
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const loadRedGoods = async () => {
```
33Orange, line 17: Same here

src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
+    const goods = await getRedGoods();
 
-    <GoodsList goods={[]} />
-  </div>
-);
+    setProduct(goods);
+  };
+
+  const load5FirstGoods = async () => {
```
33Orange, line 23: Same here

33Orange (changes requested):
Almost done

`getAll` should have errors handling
image

PR #1594, opened 2023-06-29
add task solution

Yuliia-kucheriavenko:
[DEMO LINK](https://Yuliia-kucheriavenko.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,53 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+import { Good } from './types/Good';
 import { GoodsList } from './GoodsList';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleClickGetAllGoods = async () => {
+    setGoods(await getAll());
+  };
 
-    <button type="button" data-cy="all-button">
-      Load all goods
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -5,15 +5,24 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+
+      return response.json();
+    });
 }
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((good1, good2) => good1.name.localeCompare(good2.name))
```
dffUqp, line 20: bad practice to use a number in variable naming, I think we can use `a` and `b` in sort method.

src/GoodsList.tsx
```diff
@@ -5,12 +5,16 @@ type Props = {
...
 
-export const GoodsList: React.FC<Props> = ({ goods }) => (
+export const GoodsList: React.FC<Props> = React.memo(({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li
+        key={good.id}
+        data-cy="good"
+        style={{ color: good.color }}
```
dffUqp, line 14: let's create class names for colors and add class names from `good.color`

src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
-    <h1>Dynamic list of Goods</h1>
+  const handleClickGetAllGoods = async () => {
+    setGoods(await getAll());
+  };
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleClickGet5First = async () => {
+    setGoods(await get5First());
```
dffUqp, line 15: What will we do if `get5First` return an error?

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

PR #1595, opened 2023-06-29
add dynamic list-of-goods

vchervinko:
[DEMO LINK](https://vchervinko.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,49 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleClick = async (func: () => Promise<Good[]>) => {
+    func()
```
naziamoff, line 14: wtf? (read: what's the func?)

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
+        Load all goods
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={() => handleClick(get5First)}
```
naziamoff, line 33: better to create a separate handler for each case. Reusability is great, but what if at some point in the future you gonna have to do some additional actions while loading only 5 first?

src/App.tsx
```diff
@@ -1,27 +1,49 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 // or
 // import * as goodsAPI from './api/goods';
```
naziamoff, line 8: never leave redundant comments - they are really confusing for those who read them

naziamoff (changes requested):
not bad, but still there are ways to improve your code

PR #1599, opened 2023-06-29
dynamic list of goods

YeMyr:
[DEMO LINK](https://yemyr.github.io/react_dynamic-list-of-goods/)


roman-mirzoian (changes requested):
The application is well written and works well, but to check how exactly it works, you need to download it to your computer and watch it locally, because now the link to the demo opens a Readme file. You need to figure out why the demo doesn't work and it will be perfect :)

PR #1600, opened 2023-06-29
solution

SofiaLiubchenko:
[DEMO LINK](https://SofiaLiubchenko.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,27 +1,66 @@
...
-      Load all goods
-    </button>
+  const handleAllGoods = () => {
+    getAll().then(goods => setSortedGoods(goods));
+  };
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  const handle5Goods = () => {
```
naziamoff, line 17: we usually wrap handlers like that into the `useCallback` hook. This way you have it now - we will create this hook anew on each render. With `useCallback` we won't. Not critical though, that's just a suggestion for the future)

naziamoff (approved):
awesome 🔥

PR #1601, opened 2023-07-01
Solution

Ssviatt:
[DEMO LINK](https://Ssviatt.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,63 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goodsList, setGoodsList] = useState<{
+    all: Good[],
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
+export const App: React.FC = () => {
+  const [goodsList, setGoodsList] = useState<{
+    all: Good[],
+    getFirst: Good[],
+    getRed: Good[],
+  }>({
+    all: [],
+    getFirst: [],
+    getRed: [],
+  });
```
etojeDenys, line 17: we can just create one list and make functions to change this list

```
 const [goods, setGoods] = useState<Good[]>([]);
  const setAllGoods = async () => {
    setGoods(await getAll());
  };

  const set5First = async () => {
    setGoods(await get5First());
  };

  const setRedGoods = async () => {
    setGoods(await getRedGoods());
  };
```

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,62 @@
...
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .slice(0, 5);
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const getRed = () => {
+    return [...goodsList].filter(el => el.color === 'red');
+  };
```
DarkMistyRoom, line 25: the purpose of this task is to send a new request with each click on a button

DarkMistyRoom: I see what you're doing, but this task requires another approach

DarkMistyRoom (changes requested):

PR #1602, opened 2023-07-02
dynamic list of goods was created

Pavlo-Petrashevskyi:
[DEMO LINK](https://Pavlo-Petrashevskyi.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/types/Good.ts
+++ b/src/types/Good.ts
@@ -1,4 +1,5 @@
 export interface Good {
+  localeCompare(good2: Good): number;
   id: number;
   name: string;
   color: string;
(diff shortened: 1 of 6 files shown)


src/types/Good.ts
```diff
@@ -1,4 +1,5 @@
 export interface Good {
+  localeCompare(good2: Good): number;
```
etojeDenys, line 2: why do we need this method

etojeDenys (changes requested):

src/types/Good.ts
Pavlo-Petrashevskyi: removed

SanyaBratashchuk (approved):

PR #1611, opened 2023-07-08
add task solution

ollavka:
[DEMO LINK](https://ollavka.github.io/react_dynamic-list-of-goods/)


src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,61 @@
+import { FC, memo } from 'react';
+import { Good } from '../../types/Good';
+
+function areGoodsEqual(prevGoods: Good[], currGoods: Good[]): boolean {
```
anstsot, line 4: it's better to create `utils` folder and separate `areGoodsEqual.js` file for this function

src/components/GoodsList/GoodsList.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  return true;
+}
+
+type Props = {
+  goods: Good[]
+};
+
+export const GoodsList: FC<Props> = memo(({ goods }) => (
```
anstsot, line 31: no need to use `memo`

anstsot (approved):
approving:sparkles:
But pay attention to my comments to avoid the same mistakes in the future

PR #1613, opened 2023-07-11
Solution

yura-enttsel:
[DEMO LINK](https://yura-enttsel.github.io/react_dynamic-list-of-goods/)
Demo not working, trying to fix it

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -2,10 +2,10 @@
   color: #f00;
 }
 
-.blue {
-  color: #00f;
-}
-
 .green {
   color: #008000;
+}
+
+.blue {
+  color: #00f;
 }


volodymyr-soltys97 (changes requested):
We are communicate in fe_chat

src/api/goods.ts
```diff
@@ -3,17 +3,30 @@ import { Good } from '../types/Good';
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll(): Promise<Good[]> {
+export function getData<T>(): Promise<T> {
```
volodymyr-soltys97, line 6: In the getData function, you are not specifying what type of data you are expecting as a response. This can potentially lead to type errors. Consider replacing `<T>` with `<Good[]>`.

volodymyr-soltys97 (approved):

PR #1615, opened 2023-07-12
solution

anna-akhalmosulishvili:
https://anna-akhalmosulishvili.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,65 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { get5First, getAll, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,65 @@
...
+        data-cy="all-button"
+        onClick={handleAll}
+      >
+        Load all goods
+      </button>
+
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={handle5}
```
loralevitska, line 49:
```suggestion
        onClick={handleFive}
```

loralevitska (changes requested):
Almost done) Let's improve

use `good.color` 
image
like this 
image

src/App.tsx
```diff
@@ -1,27 +1,58 @@
...
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  useEffect(() => {
+    getAll().then();
+  }, []);
```
ivbutsykin, line 13: No need in this request. It's make sense if you set state in this `useEffect` but you don't do it

ivbutsykin (approved):
Good job!

PR #1617, opened 2023-07-12
task done

pos1bl:
https://pos1bl.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,57 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [errorMessage, setErrorMessage] = useState('');
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleGoods = (getGoods: Promise<Good[]>) => {
+    getGoods
+      .then(setGoods)
+      .catch(() => setErrorMessage('Try again later'));
+  };
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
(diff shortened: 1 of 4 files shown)


anastasiiavorobiova (changes requested):
Great start! Added comments to fix

src/components/Loader/Loader.scss
```diff
@@ -0,0 +1,21 @@
+.lds-ring {
+  box-sizing: border-box;
```
OleziO, line 2: Move to * selector and remove from all other places.

OleziO (changes requested):

src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      {isLoading && <Loader />}
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      {!isLoading && (errorMessage ? (
```
polosanya, line 53: It would be great to create a variable for this condition

src/api/goods.ts
```diff
@@ -10,10 +11,12 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((good1, good2) => good1.name.localeCompare(good2.name))
```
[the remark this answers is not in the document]
polosanya: Not fixed. Create a copy of goods before sorting

src/api/goods.ts
```diff
@@ -2,18 +2,27 @@ import { Good } from '../types/Good';
...
+      .slice(0, 5));
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => [...goods].filter(good => good.color === COLOR_RED));
```
polosanya, line 27: Conversely, here the copy is redundant because the filter is a non-mutating method

polosanya (changes requested):
Almost perfect!

polosanya (changes requested):
Comments above

src/App.tsx
```diff
@@ -1,27 +1,72 @@
...
+    getGoods()
+      .then(setGoods)
+      .catch(() => setErrorMessage('Try again later'))
+      .finally(() => setIsLoading(false));
+  };
+
+  const renderContent = () => {
+    if (isLoading) {
+      return <></>;
+    }
```
vlad-demchuk, line 26: why do you need an empty jsx during the loading?
just show loading or GoodsList in jsx

also, check error below returns error message if there is some error, so it's okay

vlad-demchuk (changes requested):

volodymyr-soltys97 (approved):
Nice work 🔥

PR #1623, opened 2023-07-17
add task solution

VolodymyrPst:
https://volodymyrpst.github.io/react_dynamic-list-of-goods/


sTorba24 (changes requested):
Almost good! Style elements with their color(text colorized should be)

loralevitska (approved):
Well done)

PR #1626, opened 2023-07-18
Solution

adv555:
**[DEMO LINK](https://adv555.github.io/react_dynamic-list-of-goods/)**

### DIFF
--- a/src/GoodsList.tsx
+++ b/src/GoodsList.tsx
@@ -5,12 +5,18 @@ type Props = {
   goods: Good[]
 };
 
-export const GoodsList: React.FC<Props> = ({ goods }) => (
-  <ul>
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
-        {good.name}
-      </li>
-    ))}
-  </ul>
-);
+export const GoodsList: React.FC<Props> = ({ goods }) => {
+  return (
+    <ul>
+      {goods.map(good => (
+        <li
+          key={good.id}
+          data-cy="good"
+          style={{ color: good.color }}
+        >
+          {good.name}
+        </li>
+      ))}
+    </ul>
+  );
+};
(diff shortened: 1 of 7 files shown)


src/GoodsList.tsx
```diff
@@ -5,12 +5,18 @@ type Props = {
...
-export const GoodsList: React.FC<Props> = ({ goods }) => (
-  <ul>
-    {goods.map(good => (
-      <li key={good.id} data-cy="good">
-        {good.name}
-      </li>
-    ))}
-  </ul>
-);
+export const GoodsList: React.FC<Props> = ({ goods }) => {
```
GGsmou, line 8: It might be better to use `React.memo` here for optimisation

src/api/goods.ts
```diff
@@ -3,17 +3,22 @@ import { Good } from '../types/Good';
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll(): Promise<Good[]> {
-  return fetch(API_URL)
-    .then(response => response.json());
+export async function getAll(): Promise<Good[]> {
+  const response = await fetch(API_URL);
```
GGsmou, line 7: I think you need to check if response status is ok like this before returning it

if (!response.ok) {
...
}

GGsmou (approved):
GJ, approving
Check some moments below

src/api/goods.ts
adv555: I'm not sure about that🤔. 
the result of this async function (promise) is used in external code further and errors are processed by callback method catch(). And as I understood the error parameter value in the catch() method is an error which will be await if the promo is rejected.

anstsot (approved):
Good work, approving:sparkles:

PR #1631, opened 2023-07-19
Develop

artkas03:
[DEMO LINK](https://artkas03.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -4,8 +4,6 @@ import { GoodsList } from './GoodsList';
 
 import { get5First, getAll, getRedGoods } from './api/goods';
 import { Good } from './types/Good';
-// or
-// import * as goodsAPI from './api/goods';
 
 enum FetchingGoods {
   ALL, FIRST_FIVE, RED_ONLY,
@@ -35,7 +33,7 @@ export const App: React.FC = () => {
       }
     }
 
-    dataPromise.then((goodsData: Good[]) => setGoods(goodsData));
+    dataPromise.then(setGoods);
   }, [toFetch]);
 
   return (
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,72 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { get5First, getAll, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 // or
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,75 @@
...
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [toFetch, setToFetch] = useState<FetchingGoods>(FetchingGoods.ALL);
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  useEffect(() => {
```
GGsmou, line 16: Interesting solution with `useEffect` and `toFetch` query, one problem: list renders instantly, but should be after button click. You can just change the default value of `toFetch` and add one more case to `switch` and it should fix this problem.

GGsmou (approved):
GJ, approving!
Check one problem below)

src/api/goods.ts
```diff
@@ -10,10 +10,18 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
+      goods.sort((good1, good2) => {
+        return good1.name.localeCompare(good2.name);
+      });
```
anstsot, line 16:
```suggestion
      goods.sort((good1, good2) => good1.name.localeCompare(good2.name));
```

src/api/goods.ts
```diff
@@ -10,10 +10,18 @@ export function getAll(): Promise<Good[]> {
...
+      return goods.slice(0, 5);
+    });
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => {
+      return goods.filter(good => good.color === 'red');
+    });
```
anstsot, line 26:
```suggestion
    .then(goods => goods.filter(good => good.color === 'red'));
```

src/App.tsx
anstsot, line 16: let's not use `useEffect` hook for this task. try to load goods on click only. then you will have only one state variable

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your code.

And678 (changes requested):
[Checklist №1](https://github.com/mate-academy/react_dynamic-list-of-goods/blob/master/checklist.md) is not done

And678 (approved):
Nice! :fire:

PR #1633, opened 2023-07-19
sol

yurii-krymskiy:
[DEMO LINK](https://yurii-krymskiy.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/GoodsList.tsx
+++ b/src/GoodsList.tsx
@@ -8,17 +8,15 @@ type Props = {
 
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
-    {goods.map(good => (
+    {goods.map(({ id, name, color }) => (
       <li
-        key={good.id}
+        key={id}
         data-cy="good"
         className={cn({
-          'is-red': good.color === 'red',
-          'is-green': good.color === 'green',
-          'is-blue': good.color === 'blue',
+          [`is-${color}`]: color,
         })}
       >
-        {good.name}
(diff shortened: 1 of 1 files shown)


src/GoodsList.tsx
```diff
@@ -8,7 +9,15 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li
+        key={good.id}
+        data-cy="good"
```
volodymyr-soltys97, line 14: Instead of hardcoding the color classes ('is-red', 'is-green', 'is-blue'), consider using dynamic class name based on the color of the good. This will make your code more scalable when there are more colors available.

src/GoodsList.tsx
```diff
@@ -8,7 +9,15 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
```
volodymyr-soltys97, line 11: It will be better if you use a desctructurization for good here

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

VitaliyBondarenko1982 (changes requested):
Please fix prev comments

volodymyr-soltys97 (approved):
Well done 👍

PR #1647, opened 2023-07-24
Solution

outofs:
https://outofs.github.io/react_dynamic-list-of-goods/


src/App.tsx
```diff
@@ -1,27 +1,42 @@
...
+        Load all goods
+      </button>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={() => get5First().then(setGoods)}
```
eleonora-demikh, line 26: same here

src/App.tsx
```diff
@@ -1,27 +1,42 @@
...
+        Load 5 first goods
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={() => getRedGoods().then(setGoods)}
```
eleonora-demikh, line 34: same here

src/App.tsx
```diff
@@ -1,27 +1,42 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => getAll().then(setGoods)}
```
eleonora-demikh, line 18: It is better to carry out this expression in separate functions, and not prescribed in onClick

src/api/goods.ts
```diff
@@ -10,10 +10,18 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods
+      .sort((goodA: Good, goodB: Good) => goodA.name.localeCompare(goodB.name))
+      .slice(0, 5))
+    .catch(error => {
```
eleonora-demikh, line 16: If you add error processing, then it must be carried out in a separate ```if ``` with ```return``` before processing the received data

eleonora-demikh (changes requested):
Good job! Here are some small fixes

And678 (changes requested):
[Checklist №1](https://github.com/mate-academy/react_dynamic-list-of-goods/blob/master/checklist.md) is not satisfied

And678 (approved):
:+1: :+1:

PR #1652, opened 2023-07-24
React_dynamic-list-of-goods

TamaraShukalo:
[DEMO LINK](https://TamaraShukalo.github.io/react_dynamic-list-of-goods/)


src/GoodsList.tsx
```diff
@@ -8,7 +8,11 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
```
anstsot, line 8: Make sure that you add some rendering optimization techniques (React.memo and so on.)

anstsot (changes requested):
Let's add memoization

PR #1659, opened 2023-07-25
solution

Valentyna08:
[DEMO LINK](https://Valentyna08.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,60 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,60 @@
...
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleLoadAllGoods = () => {
+    getAll().then((data) => {
+      setGoods(data);
+    });
+  };
```
NataliGru, line 15: In my view, it'll be better to create one general function, like handleClick, which takes a getData function as a parameter, because the logic of actions remains the same every time)

NataliGru (approved):
Looks good!))

anstsot (changes requested):
[Checklist #3](https://github.com/mate-academy/react_dynamic-list-of-goods/blob/master/checklist.md) is not satisfied

src/App.tsx
YegorKorenkov, line 15: You can memoize todos so not to request from each time. For example:
`
const allTodos = useMemo(() => getAll(), [])
`
And the same for each request

src/GoodsList.tsx
```diff
@@ -5,12 +5,16 @@ type Props = {
   goods: Good[]
 };
 
-export const GoodsList: React.FC<Props> = ({ goods }) => (
+export const GoodsList: React.FC<Props> = React.memo(({ goods }) => (
```
YegorKorenkov, line 8: React memo redundant here, because goods changed every time then you click on the button(getAll, get5...)

YegorKorenkov (approved):
Approve, but please pay attention on comments

PR #1663, opened 2023-07-25
solution

mariuszgit:


src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleAllGoods = () => {
```
Radoslaw-Czerniawski, line 10: Generally all functions and variables (except hooks) on every re-render of the component (for example when state changes or components unmounts and mounts again) are again redeclared. We can mitigate that by using memoization hooks like [useCallback](https://react.dev/reference/react/useCallback) and [useMemo](https://react.dev/reference/react/useMemo). 
Here it is not a big burden for the apps performance, however in bigger project it's good to apply them. Let's imagine a scenario where you return a component in the jsx that accepts a handler like here, if the handler was not memoized, component that accepts it would rerender on every state change of the parent (reference in props changes -> props change -> component rerenders)

Radoslaw-Czerniawski: Also all of these api calls lack any kind of error catching. If one of these api calls fails your app will throw an error and app will die.
Either apply [.catch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) or use async functions with [try catch blocks](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch)

Radoslaw-Czerniawski: Also for the future prs (here it is applied and that's good) if it's fitting try to apply the [presentational-container](https://www.patterns.dev/posts/presentational-container-pattern) pattern.

Radoslaw-Czerniawski (changes requested):
All good apart from lack of the error catching.

DorotaLeniecDev (approved):
Good job

PR #1666, opened 2023-07-25
add task solution

klmyana:


src/App.tsx
```diff
@@ -1,27 +1,75 @@
...
-    <GoodsList goods={[]} />
-  </div>
-);
+export const App: React.FC = () => {
+  const [goodList, setGoodList] = useState<Good[]>([]);
+  const [filter, setFilter] = useState<ListFilterType>(ListFilterType.None);
+
+  useEffect(() => {
+    getFiltredList(filter)?.then(products => setGoodList(products));
+  }, [filter]);
```
anstsot, line 36: try to remove `filter` state variable and use `useCallback` hook to load goods on click

anstsot (changes requested):
[Checklist #1 and #3](https://github.com/mate-academy/react_dynamic-list-of-goods/blob/master/checklist.md) are not satisfied

src/GoodsList.tsx
```diff
@@ -5,12 +5,16 @@ type Props = {
   goods: Good[]
 };
 
-export const GoodsList: React.FC<Props> = ({ goods }) => (
+export const GoodsList: React.FC<Props> = React.memo(({ goods }) => (
```
YegorKorenkov, line 8: React memo redundant here, because goods changed every time when you click on the button(getAll, get5...)

src/App.tsx
```diff
@@ -1,27 +1,53 @@
...
-// import * as goodsAPI from './api/goods';
+export const App: React.FC = () => {
+  const [goodList, setGoodList] = useState<Good[]>([]);
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+  const handleGetAll = () => {
+    return getAll().then(setGoodList);
+  };
```
YegorKorenkov, line 12: You can memoize todos so not to request from each time. For example:
const allTodos = useMemo(() => getAll(), [])
And the same for each request

YegorKorenkov (approved):
Good job! please pay attention on comments

PR #1671, opened 2023-07-25
Solution

perfectAcedia:
[DEMO LINK](https://perfectAcedia.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,75 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+const enum Actions {
+  AllGoods = 'All goods',
+  FirstFive = 'first-five',
+  OnlyRed = 'red',
+}
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState<Good[]>([]);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,75 @@
...
+          .then(goodsFromServer => {
+            setVisibleGoods(goodsFromServer);
+          });
 
-    <GoodsList goods={[]} />
-  </div>
-);
+        break;
+    }
+  }, [handleAction]);
```
anstsot, line 42: rewrite the logic by using `useCallback` hook

anstsot (changes requested):
[Checklist #1 and #3](https://github.com/mate-academy/react_dynamic-list-of-goods/blob/master/checklist.md) are not satisfied

src/App.tsx
```diff
@@ -1,27 +1,79 @@
...
+  FirstFive = 'first-five',
+  OnlyRed = 'red',
+}
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [visibleGoods, setVisibleGoods] = useState<Good[]>([]);
+  const [handleAction, setHandleAction] = useState(Action.AllGoods);
```
anstsot, line 15: the `handleAction` state variable isn't necessary here. you can just call the function on click and set visibleGoods

anstsot (changes requested):
There shouldn't be any goods by default. Show them after click only
image

src/App.tsx
```diff
@@ -1,27 +1,64 @@
...
+        setVisibleGoods(goodsFromServer);
+      });
+  }, []);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  useEffect(() => {
+    setVisibleGoods([]);
+  }, []);
```
anstsot, line 31: redundant
```suggestion
```

anstsot (approved):
Good work, approving:sparkles:

PR #1690, opened 2023-07-29
solution

vlkzmn:
[DEMO LINK](https://vlkzmn.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,42 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,42 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => getAll().then(setGoods)}
```
Anastasiia-Svintsova, line 18: You repeat the same logic for each onClick action. Try to refactor it, to not repeat this part `.then(setGoods)`. You can create a separate function, for example

src/GoodsList.tsx
```diff
@@ -8,7 +8,11 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
-      <li key={good.id} data-cy="good">
+      <li
+        key={good.id}
```
Anastasiia-Svintsova, line 12: Use destructuring for the good object for a better readability

Anastasiia-Svintsova (changes requested):
Good start! Check my comments below

maxim2310 (approved):
Nice work!!!

PR #1698, opened 2023-08-02
solution

Krivets:
[DEMO LINK](https://Krivets.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,47 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,47 @@
...
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => {
+          getAll().then(setGoodsList);
+        }}
```
Anastasiia-Svintsova, line 19: You repeat the same logic (.then(setGoodsList)` for each button. Create a separate method for it

Anastasiia-Svintsova (changes requested):
Almost done!

PR #1713, opened 2023-08-14
task solution

Bonum-taurus:
https://Bonum-taurus.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,44 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [filteredGoods, setfilteredGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  return (
+    <div className="App">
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,44 @@
...
-    </button>
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        onClick={() => getAll().then(setfilteredGoods)}
```
anastasiiavorobiova, line 18: The promises returned by the API calls need error handling. You should add a catch block to handle any errors that may occur during the API call.

anastasiiavorobiova (changes requested):
Great work! Added comments to fix

src/App.tsx
```diff
@@ -1,27 +1,55 @@
...
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        onClick={() => getAll()
+          .then(setFilteredGoods)
+          .catch(error => setErrorMessage(error.message))}
+        type="button"
```
volodymyr-soltys97, line 22: It is not a very good idea to write logic like this in `jsx`, it is better to create auxiliary functions and use them here

src/App.tsx
```diff
@@ -1,27 +1,55 @@
...
+        Load all goods
+      </button>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        onClick={() => get5First()
+          .then(setFilteredGoods)
+          .catch(error => setErrorMessage(error.message))}
```
volodymyr-soltys97, line 31: The same as above

volodymyr-soltys97 (changes requested):
Let's improve your code

anastasiiavorobiova (approved):
Looks amazing!

PR #1715, opened 2023-08-18
Solution

s3rg4sh:
https://s3rg4sh.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/api/goods.ts
+++ b/src/api/goods.ts
@@ -5,15 +5,28 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
 
 export function getAll(): Promise<Good[]> {
   return fetch(API_URL)
-    .then(response => response.json());
+    .then(response => {
+      if (!response.ok) {
+        throw new Error('Not your day... Please try again later.');
+      }
+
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -5,15 +5,28 @@ const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/good
...
+
+      return sortedGoods.splice(0, 5);
+    });
 };
 
-export const getRedGoods = () => {
+export const getRedGoods = (): Promise<Good[]> => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => goods)
```
etojeDenys, line 30: looks like redundant then

etojeDenys (changes requested):

src/api/goods.ts
lerastarynets: @s3rg4sh remove this `then` or explain why u need it pls

lerastarynets (changes requested):
Pls fix prev comments

witflash (approved):
Good job 👍🏻

PR #1718, opened 2023-08-21
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,45 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+// import { getAll, get5First } from './api/goods';
+import { Good } from './types/Good';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [allGoods, setAllGoods] = useState<Good[] | null>(null);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={async () => setAllGoods(await getAll())}
+      >
+        Load all goods
+      </button>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={async () => setAllGoods(await get5First())}
+      >
+        Load 5 first goods
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={async () => setAllGoods(await getRedGoods())}
+      >
+        Load red goods
+      </button>
+
+      <GoodsList goods={allGoods || []} />
+    </div>
+  );
+};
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,45 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={async () => setAllGoods(await getAll())}
```
darokrk, line 21: Avoid using async functions in the onClick handler. Instead, create a separate async function and call that function in the onClick handler e.g. add this before return

```
  const showAllGoods = async () => {
    setGoods(await getAll());
  };
```

```
<button
  type="button"
  data-cy="all-button"
  onClick={showAllGoods}
>
```

src/App.tsx
```diff
@@ -1,27 +1,45 @@
...
+        Load all goods
+      </button>
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <button
+        type="button"
+        data-cy="first-five-button"
+        onClick={async () => setAllGoods(await get5First())}
```
darokrk, line 29: Same here, create separate function e.g.

```
  const showFiveGoods = async () => {
    setGoods(await get5First());
  };
```

src/App.tsx
```diff
@@ -1,27 +1,45 @@
...
+        Load 5 first goods
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={async () => setAllGoods(await getRedGoods())}
```
darokrk, line 37: Same here, e.g.:

```
  const showRedGoods = async () => {
    setGoods(await getRedGoods());
  };
```

src/api/goods.ts
```diff
@@ -3,17 +3,29 @@ import { Good } from '../types/Good';
 // eslint-disable-next-line
 const API_URL = `https://mate-academy.github.io/react_dynamic-list-of-goods/goods.json`;
 
-export function getAll(): Promise<Good[]> {
+export function getAll(): Promise<Good[] | null> {
```
darokrk, line 6: The type of the Promise returned by the getAll function has been changed from Good[] to Good[] | null. This might not be necessary as fetch API doesn't return null but throws an error in case of failure.

```
return fetch(API_URL)
    .then(response => {
      if (!response.ok) {
        throw new Error(`${response.status} ${response.statusText}`);
      }

      return response.json();
    })
```

src/App.tsx
```diff
@@ -1,27 +1,45 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+// import { getAll, get5First } from './api/goods';
```
darokrk, line 6: remove commented code

src/api/goods.ts
```diff
@@ -3,17 +3,29 @@ import { Good } from '../types/Good';
...
   return fetch(API_URL)
     .then(response => response.json());
 }
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => {
+      if (!goods) {
```
darokrk, line 14: then if you handled Error check in `getAll()` method you can remove this null check

src/api/goods.ts
```diff
@@ -3,17 +3,29 @@ import { Good } from '../types/Good';
...
+
+      return goods.sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5);
+    }); // sort and get the first 5
 };
 
 export const getRedGoods = () => {
   return getAll()
-    .then(goods => goods); // get only red
+    .then(goods => {
+      if (!goods) {
```
darokrk, line 25: same here

darokrk (changes requested):
Looks good 👏🏽  few mentions left for improvements 🛠️

Radoslaw-Czerniawski (approved):
Approved :D

PR #1719, opened 2023-08-21
first commit

haduigon:
https://haduigon.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,49 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
-
-// import { getAll, get5First, getRed } from './api/goods';
+import { Good } from './types/Good';
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState([] as Good[]);
+  /* eslint-disable */
+  function setVisibleGoods(callback: any) {
+    callback().then((res: Good[]) => { setGoods(res) }
+    )
+  }
+  /* eslint-enable */
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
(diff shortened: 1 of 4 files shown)


src/api/goods.ts
```diff
@@ -10,10 +22,10 @@ export function getAll(): Promise<Good[]> {
 
 export const get5First = () => {
   return getAll()
-    .then(goods => goods); // sort and get the first 5
+    .then(goods => goods.sort(compare).splice(0, 5)); // sort and get the first 5
```
lerastarynets, line 25: You are using the splice method on the array returned by sort. This will alter the original array. It's better to use slice which does not modify the original array.

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState([] as Good[]);
+  /* eslint-disable */
+  function setVisibleGoods(callback: any) {
+    callback().then((res: Good[]) => { setGoods(res) }
```
lerastarynets, line 13: The Promise returned by 'callback()' function needs error handling. Please use .catch or try/catch to handle potential errors.

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState([] as Good[]);
+  /* eslint-disable */
```
lerastarynets, line 11: Avoid disabling eslint if possible. It's better to address the issues that eslint is flagging, rather than ignoring them.

src/App.tsx
```diff
@@ -1,27 +1,49 @@
...
+  function setVisibleGoods(callback: any) {
```
lerastarynets, line 12: Avoid using 'any' as a type declaration. It defeats the purpose of using TypeScript because it does not provide any type safety.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)
Also pls consider adding css color property to list items

src/api/goods.ts
haduigon: done

src/App.tsx
haduigon: done

TarasHoliuk (approved):

PR #1727, opened 2023-08-28
task solution

oleksandrshtonda:
[DEMO LINK](https://oleksandrshtonda.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,68 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+enum FUNCS {
+  GET_ALL = 'getAll',
+  GET_FIRST_5 = 'getFirstFive',
+  GET_RED = 'getRed',
+}
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,68 @@
...
+  GET_FIRST_5 = 'getFirstFive',
+  GET_RED = 'getRed',
+}
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [functionGetData, setFunctionGetData] = useState(FUNCS.GET_ALL);
```
Anastasiia-Svintsova, line 16: Do you really need this state for function data? You can remove it and refactor the code

src/App.tsx
```diff
@@ -1,27 +1,68 @@
...
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+  useEffect(() => {
+    if (functionGetData === FUNCS.GET_RED) {
```
Anastasiia-Svintsova, line 19: You don't need to call these functions inside the useEffect. Just call the correct function on the button click

src/App.tsx
```diff
@@ -1,27 +1,68 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => setFunctionGetData(FUNCS.GET_ALL)}
```
Anastasiia-Svintsova, line 44: You can create some function for setting goods and pass the callback there. Then just call the callback inside this function and set the goods to the goods state. The logic is too complicated now

Anastasiia-Svintsova (changes requested):
Use correct API
image
image
image

volodymyr-soltys97 (approved):
Well done 🔥

PR #1735, opened 2023-09-05
solution

tndeineka:
[DEMO LINK](https://tndeineka.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,61 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,27 +1,61 @@
...
+    if (status === Status.First5) {
+      get5First().then(setGoods);
+    }
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+    if (status === Status.OnlyRed) {
+      getRedGoods().then(setGoods);
+    }
```
VitaliyBondarenko1982, line 27: better to use switch case instead three if

VitaliyBondarenko1982 (changes requested):

IvanFesenko (approved):
Well done

PR #1741, opened 2023-09-10
list of goods

MaksymKos:
https://maksymkos.github.io/react_dynamic-list-of-goods/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,44 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,27 +1,44 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={async () => setGoods(await getAll())}
```
VitaliyBondarenko1982, line 20: better to make one function factory and use it for all buttons
```
   const handleGoods = (callback) => async () => setGoods(await callbacl())
```

```suggestion
        onClick={handleGoods(getAll)}
```

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comment

etojeDenys (approved):
good job

PR #1743, opened 2023-09-11
solution

HYPER2307:
[DEMO LINK](https://hyper2307.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,63 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
+import { get5First, getAll, getRedGoods } from './api/goods';
 
 // import { getAll, get5First, getRed } from './api/goods';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
-  </div>
-);
+    setGoods(fiveFirstGoods);
+  }
+
+  async function shouldShowRed() {
+    const redGoods = await getRedGoods();
+
+    setGoods(redGoods);
+  }
```
VitaliyBondarenko1982, line 30: instead of three function you can create one function factory and pass to it proper callback for each button. 
```
const handleGoods = (callback) => async () => setGoods(await callback());
```

src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
+
+  return (
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
+
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={shouldShowAllGoods}
```
VitaliyBondarenko1982, line 39:
```suggestion
        onClick={handleGoods(getAll)}
```
same for other buttons

VitaliyBondarenko1982 (changes requested):
Good job
Please check comments

VitaliyBondarenko1982 (approved):
Well done

PR #1850, opened 2023-10-12
Dev

donizer:
[DEMO LINK](https://donizer.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/GoodsList.tsx
+++ b/src/GoodsList.tsx
@@ -2,18 +2,20 @@ import React from 'react';
 import { Good } from './types/Good';
 
 type Props = {
-  goods: Good[]
+  goods: Good[];
 };
 
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
-    {goods.map(good => {
-      const style: React.CSSProperties = {
-        color: good.color,
-      };
-
+    {goods.map((good) => {
       return (
-        <li key={good.id} data-cy="good" style={style}>
+        <li
+          key={good.id}
+          data-cy="good"
+          style={{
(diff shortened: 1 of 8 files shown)


donizer:
There's error on tests. I fixed it locally executing `npm i` two times. Don't know what to do here.

src/index.tsx
```diff
@@ -1,6 +1,9 @@
+import React from 'react';
 import ReactDOM from 'react-dom';
 import { App } from './App';
 
+React.createElement('div');
```
VitaliyBodnar, line 5: why do you need that ?
In my opinion it's redundant

src/GoodsList.tsx
```diff
@@ -7,10 +7,16 @@ type Props = {
...
-    ))}
+    {goods.map(good => {
+      const style: React.CSSProperties = {
+        color: good.color,
+      };
+
+      return (
+        <li key={good.id} data-cy="good" style={style}>
+          {good.name}
+        </li>
```
VitaliyBodnar, line 18: don't see any sense to create a separate object for only one style 
```suggestion
      return (
        <li 
           key={good.id} 
           data-cy="good" 
           style={{ color: good.color }}
          >
           {good.name}
        </li>
```

VitaliyBodnar (changes requested):

src/index.tsx
donizer: It throws an error without importing React or if I import React but don't call it.

donizer: image| {width:100%}

From different task, but same error.

PR #1941, opened 2023-11-20
stage-1

no4kar:
[DEMO LINK](https://no4kar.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,12 +3,16 @@ import './App.scss';
 import { GoodsList } from './GoodsList';
 import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 // or
-import * as goodsAPI from './api/goods';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,27 +1,44 @@
...
+    <div className="App">
+      <h1>Dynamic list of Goods</h1>
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      <button
+        type="button"
+        data-cy="all-button"
+        onClick={() => goodsAPI.getAll().then(setGoods)}
```
Esceype, line 20: Move all your actions for buttons to the handler, it can be one handler for all there is only one difference.

Esceype (changes requested):
GJ! Left just one comment.

src/App.tsx
no4kar: if corrections are right, could you leave opinion about how would be better named this handler. thanks.

PR #2771, opened 2025-03-24
solution

sansamiste:
[DEMO LINK](https://sansamiste.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,64 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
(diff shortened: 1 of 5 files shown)


PiekharievaK (approved):
good job.

src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
+          ? err.message
+          : 'Failed to load goods. Please try again.',
+      );
+    }
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const visibleGoods = useMemo(() => goods, [goods]);
```
alexander-ignatow, line 28: This memoization does nothing.

It recalculates each time goods change.

src/App.tsx
```diff
@@ -1,27 +1,63 @@
...
+import React, { useMemo, useState } from 'react';
 import './App.scss';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+const MemoizedGoodsList = React.memo(GoodsList);
```
alexander-ignatow, line 7: This memoization works as expected, but most probably you won't notice this in your app

alexander-ignatow (changes requested):
Looks good, but please, fix memoization usage issues

src/App.tsx
sansamiste: I understand that there is no need to use it here, the AI ​​just requested optimization via React.memo, thank you for your attention!

alexander-ignatow (approved):
👍

PR #2773, opened 2025-03-24
add task solution

AnOvramenko:
[DEMO LINK](https://AnOvramenko.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,54 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 // or
 // import * as goodsAPI from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,54 @@
...
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleAllGoods = () => {
+    getAll().then((goodsFromServer: Good[]) => {
```
daniil-trusov, line 14: getAll.then will always return Good[], so it's possible to simplify code there:
`getAll().then(setGoods);`

src/App.tsx
```diff
@@ -1,27 +1,54 @@
...
+      setGoods(goodsFromServer);
+    });
```
daniil-trusov, line 16: It's also a good practive to catch possible errors after then():
```
.then(setGoods)
.catch(() => setErrorMessage('Something went wrong')),
```

Error message can be shown instead of goodlist in case of errors.

daniil-trusov (approved):
Good job! Please check comments to see how you can improve or simplify your solution.

PR #2778, opened 2025-03-24
solution

Mazafakerqq:
[DEMO LINK](https://Mazafakerqq.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,42 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
+import { Good } from './types/Good';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
 
-export const App: React.FC = () => (
-  <div className="App">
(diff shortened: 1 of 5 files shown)


src/api/goods.ts
```diff
@@ -8,9 +8,11 @@ export function getAll(): Promise<Good[]> {
 }
 
 export const get5First = () => {
-  return getAll().then(goods => goods); // sort and get the first 5
+  return getAll().then(goods =>
+    goods.sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5),
```
radykhovska, line 12: it's better to give full names for variables

src/App.tsx
```diff
@@ -1,27 +1,42 @@
...
-    </button>
+  const load5First = () => {
+    get5First().then(setGoods);
+  };
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+  const redGoods = () => {
+    getRedGoods().then(setGoods);
```
radykhovska, line 20: you should handle possible errors during the fetch operation

radykhovska (approved):
good job! left several comments

src/api/goods.ts
alexander-ignatow: @radykhovska  you're right, but IMO in this case all is fine )

PR #2780, opened 2025-03-24
Solution

fromjkeee16:
[DEMO LINK](https://fromjkeee16.github.io/react_dynamic-list-of-goods/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,27 +1,66 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
 import './App.scss';
 import { GoodsList } from './GoodsList';
 
-// import { getAll, get5First, getRed } from './api/goods';
-// or
-// import * as goodsAPI from './api/goods';
+import { getAll, get5First, getRedGoods } from './api/goods';
+import { Good } from './types/Good';
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+export const App: React.FC = () => {
+  const [goods, setGoods] = useState<Good[]>([]);
+  const [errorMsg, setErrorMsg] = useState('');
 
-    <button type="button" data-cy="all-button">
-      Load all goods
-    </button>
+  const handleGoodsSelect = async (promise: Promise<Good[]>) => {
+    try {
+      const fetchedGoods = await promise;
 
-    <button type="button" data-cy="first-five-button">
-      Load 5 first goods
-    </button>
+      setGoods(fetchedGoods);
+      setErrorMsg('');
+    } catch (error) {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,27 +1,66 @@
...
+      {errorMsg && <p className="help is-danger">{errorMsg}</p>}
+      {!errorMsg && goods.length === 0 && (
+        <p className="help is-warning">
+          There are no goods or they weren&apos;t loaded
+        </p>
+      )}
+      {!errorMsg && goods.length > 0 && goodsComponent}
+    </div>
+  );
+};
```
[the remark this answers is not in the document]
fromjkeee16: @mateacademy-ai-mentor there's a `useMemo()` in `App.tsx` on line 29

`const goodsComponent = useMemo(() => <GoodsList goods={goods} />, [goods]);`

src/App.tsx
```diff
@@ -1,27 +1,68 @@
...
-      Load red goods
-    </button>
+      setGoods(fetchedGoods);
+      setErrorMsg('');
+    } catch (error) {
+      if (error instanceof Error) {
+        setErrorMsg(error.message);
+      } else {
+        setErrorMsg('An unexpected error occured');
+      }
```
alexander-ignatow, line 25: the type check here looks like an overkill, but if you still need it, I suggest following:

```suggestion
        setErrorMsg(error.message ?? 'An unexpected error occured');
```

src/App.tsx
```diff
@@ -1,27 +1,68 @@
...
+    }
+  };
 
-    <GoodsList goods={[]} />
-  </div>
-);
+  const goodsComponent = useMemo(() => <GoodsList goods={goods} />, [goods]);
```
alexander-ignatow, line 29: If you need a memoized component, you should use `React.memo`. Please, don't use the hook to memorize the component

See for instance this article for the details:
https://blog.logrocket.com/react-memo-vs-usememo/

alexander-ignatow: In your particular case, please think, about why you need the memorization at all. What are you trying to improve? What do you actually gain from it?

src/App.tsx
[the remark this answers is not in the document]
alexander-ignatow: > @mateacademy-ai-mentor there's a `useMemo()` in `App.tsx` on line 29
> 
> `const goodsComponent = useMemo(() => <GoodsList goods={goods} />, [goods]);`

You are using `useMemo`, but doing it in a wrong way.

alexander-ignatow (changes requested):
You are on the correct path, but please, but please, fix the issues in your code

alexander-ignatow (approved):
👍

PR #2841, opened 2025-05-14
add task solution

hetskoweb:
[DEMO LINK](https://hetskoweb.github.io/react_dynamic-list-of-goods/)


etojeDenys (changes requested):
1. your demo link doesn't work, try to fix that
2. try also to pass all tests and then re-request the review

etojeDenys (approved):

PR #2905, opened 2025-06-09
Solution

meryam99:
https://meryam99.github.io/react_dynamic-list-of-goods/


zbvika (approved):
Good job! But note that you're using the mutating sort method without making a copy of the array you're working with.

ArturErmolenko (approved):

PR #3061, opened 2025-09-18
solve changes

EtienneD10:
changes solved


EtienneD10 (commented):
https://EtienneD10.github.io/react_dynamic-list-of-goods/

PR #3085, opened 2025-10-03
add solution

Aiiyuu:
[DEMO LINK](https://aiiyuu.github.io/react_dynamic-list-of-goods/)


src/App.tsx
```diff
@@ -1,27 +1,99 @@
...
 
-export const App: React.FC = () => (
-  <div className="App">
-    <h1>Dynamic list of Goods</h1>
+enum GoodsFilter {
+  All = 'All',
+  FirstFive = 'FirstFive',
+  RedOnly = 'RedOnly',
+  Default = All, // I'm not sure is it valid to do so
+}
```
781calmstate, line 13: its a valid code, keep in mind both members share the same value. Enum serves as the representation of some values and in your case you intentionally duplicate the value. I wouldn't recommend to do it like this.

781calmstate (approved):

PR #3098, opened 2025-10-17
_add solution: dynamic goods list with filtering buttons

S-Khraban:
[DEMO LINK](https://S-Khraban.github.io/react_dynamic-list-of-goods/)


src/types/Good.ts
```diff
@@ -1,5 +1,5 @@
-export interface Good {
+export type Good = {
   id: number;
   name: string;
-  color: string;
-}
+  color: 'red' | 'green' | 'blue' | string;
```
781calmstate: why did you keep the string type for this property if you decided to specify the exact strings - `'red' | 'green' | 'blue'`

src/App.tsx
```diff
@@ -1,27 +1,76 @@
...
 
-    <button type="button" data-cy="red-button">
-      Load red goods
-    </button>
+      <div className="App__buttons">
+        <button
+          type="button"
+          data-cy="all-button"
+          onClick={() => handleLoadGoods(getAllGoods, 'all')}
+          className={cn('App__button', { 'App__button--active': activeFilter === 'all' })}
```
781calmstate: you should use enum if you need the value from your `Filter` type.

It's better to be able to change the values in one place. in your current implementation you would have to scroll your whole file to change the value in each place, but if you use enum, you need to change the value only where you declare enum.

The second scenario in my example seems a bit better to me:)

src/App.tsx
```diff
@@ -1,27 +1,76 @@
...
+        </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+        <button
+          type="button"
+          data-cy="first-five-button"
+          onClick={() => handleLoadGoods(getFirstFiveGoods, 'firstFive')}
+          className={cn('App__button', { 'App__button--active': activeFilter === 'firstFive' })}
```
781calmstate: same here

src/App.tsx
```diff
@@ -1,27 +1,76 @@
...
+          disabled={isLoading}
+        >
+          Load 5 first goods
+        </button>
+
+        <button
+          type="button"
+          data-cy="red-button"
+          onClick={() => handleLoadGoods(getRedGoods, 'red')}
+          className={cn('App__button', { 'App__button--active': activeFilter === 'red' })}
```
781calmstate: same here

781calmstate (changes requested):
I wonder where did you get the idea to add `_` before your commits, I've never seen anything like this in the commercial projects. My advice is to stick to the rules from [the topic about conventional commit namings.](https://www.conventionalcommits.org/en/v1.0.0/)

781calmstate (approved):
Good job!

PR #3162, opened 2025-12-11
task

iryna-knyzh:


iryna-knyzh:
@mateacademy-ai-mentor 
There's no such requirements

PR #3198, opened 2026-01-22
Solution

Br0No:
https://br0no.github.io/react_dynamic-list-of-goods/


src/api/goods.ts
```diff
@@ -8,9 +8,11 @@ export function getAll(): Promise<Good[]> {
 }
 
 export const get5First = () => {
-  return getAll().then(goods => goods); // sort and get the first 5
+  return getAll().then(goods =>
+    goods.sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5),
```
TetianaSobolieva: The sort() method mutates the array in place. Better for example: [...goods].sort(...)

TetianaSobolieva (approved):
Good Job! 
the sort() method mutates the array directly. It is better to use [...goods].sort(...) or toSorted()

PR #3203, opened 2026-01-22
added solution

Demtar23:
[DEMO LINK](https://Demtar23.github.io/react_dynamic-list-of-goods/) added fixed link


SergiyKusiy (changes requested):
The code seems to be ok, but Demo Link doesn't work! Please fix this
image

PR #3205, opened 2026-01-22
first

akhvarastyan-web:
[DEMO LINK](https://akhvarastyan-web.github.io/react_dynamic-list-of-goods/)


src/api/goods.ts
```diff
@@ -7,10 +7,14 @@ export function getAll(): Promise<Good[]> {
   return fetch(API_URL).then(response => response.json());
 }
 
-export const get5First = () => {
-  return getAll().then(goods => goods); // sort and get the first 5
+export const get5First = (): Promise<Good[]> => {
+  return getAll().then(goods => {
+    return goods.sort((a, b) => a.name.localeCompare(b.name)).slice(0, 5);
```
dmytro-halieba, line 22: The API request does not handle possible errors, which violates checklist item #3 (API). If getAll() fails, the promise rejection is not handled, and this may lead to unstable or unpredictable behavior in the application. Error handling should be added (for example, using .catch) to ensure the function always returns a safe and predictable result.

src/GoodsList.tsx
```diff
@@ -8,7 +8,7 @@ type Props = {
 export const GoodsList: React.FC<Props> = ({ goods }) => (
   <ul>
     {goods.map(good => (
```
dmytro-halieba, line 10: Additionally, the GoodsList component does not apply any rendering optimization techniques, which violates checklist item #1 (Code Knowledge). Since this component is purely driven by props and renders a list, it should be wrapped with React.memo to prevent unnecessary re-renders when the goods prop has not changed. This optimization becomes especially important as the list grows or when the parent component re-renders frequently.

dmytro-halieba (changes requested):
Good job so far! Your application works flawlessly, but I’m requesting changes due to the code violating several points in the checklist. Please fix the commented sections, and you should be able to pass it with ease.

src/GoodsList.tsx
akhvarastyan-web: thanks for the comments. If you can, please send me a link to the checklist (Knowledge of Code). I can't find such an item. I would be very grateful

PR #3308, opened 2026-04-08
Implement loading the list of goods

Dianaeva:
[DEMO LINK](https://dianaeva.github.io/react_dynamic-list-of-goods/)


polosanya (approved):

src/App.tsx
```diff
@@ -1,27 +1,97 @@
...
+import { Loader } from './components/Loader';
+
+const getData = (filter: GoodsFilter) => {
+  switch (filter) {
+    case 'all':
+      return getAll();
+    case 'first-five':
+      return get5First();
+    case 'red':
+      return getRedGoods();
```
polosanya, line 16: Don't forget to add default case. Also, better use Enum instead of hardcode cases

PR #3334, opened 2026-05-01
Complete Goodlist implementation with dinamic colors

RVDotsenko:
[DEMO LINK](https://RVDotsenko.github.io/react_dynamic-list-of-goods/)


brespect (changes requested):
Good progress, but your DEMO LINK is not working

PR #3352, opened 2026-05-12
feat: implement dynamic list of goods

linska:
[demo](https://linska.github.io/react_dynamic-list-of-goods/)


dothearts (approved):
Approved!

PR #3363, opened 2026-05-15
finished task

Evgeniy45:
DEMO - https://evgeniy45.github.io/react_dynamic-list-of-goods/


src/App.tsx
```diff
@@ -1,27 +1,42 @@
...
+        Load 5 first goods
+      </button>
 
-    <GoodsList goods={[]} />
-  </div>
-);
+      <button
+        type="button"
+        data-cy="red-button"
+        onClick={() => getRedGoods().then(setGoods)}
```
dothearts, line 34: In this case, you skip the catch case. 
For this moment, it's okay. In future cases, write a catch in all places where you can get an error.

dothearts (approved):

